Issue #8, opened 2019-05-02
Chart is not idiomatic


andor44:
This is a somewhat subjective issue title, but there are a couple things that stand out to me about the chart, having worked with many of the [official](https://github.com/helm/charts/tree/master/stable/) charts, just a couple things off the top of my mind:
1. "Hardcoded" namespace. The chart wants to create its own namespace and install things in there. I can't see a good reason this should be the case, while at the same time it has many downsides: the chart can't be used in environments where helm doesn't have cluster-wide access so namespaces can't be created, it can't be installed in an existing namespace, etc. Unless there is a strong reason a chart should be creating its own namespace I would avoid this.
2. Hardcoded to run on SSL. While encryption is not a bad default many environments terminate SSL before traffic enters k8s. Keeping SSL/port 443 as a default is ok but I would make running it on port 80 an option.
3. The [default dashboard ingress config](https://github.com/TykTechnologies/tyk-helm-chart/blob/232d8ab43a2a9fa432a0af8826600fd3f1ac59af/tyk-pro/templates/ingress-dash.yaml#L15-L16) has two issues: it has a hardcoded ingress class "nginx". The default installation of the nginx ingress controller does use that but that's a configurable parameter. It should be configurable here too. The other annotation uses the old `ingress.kubernetes.io/rewrite-target` key. That has an `nginx.` prefix since quite a while ago, so users of more recent versions of the ingress controller will have that silently ignored.

These are hard issues that IMO need fixing. The following points are more debatable but still worth mentioning:
1. Why is the gateway a DaemonSet? If it used a `hostPort` setup I would understand but this way I see no benefit to it not just being a plain old Deployment. I would make this configurable (see official nginx ingress controller chart for example).
2. Allowing a LoadBalancer type service and/or Ingresses (for different listen hosts) for the gateway would be a welcome addition.

I will add more points as they come up.

excieve:
Thanks for your valuable feedback @andor44! This chart is very much work in progress and will be improved. Basically agree on all your points.

Regarding `DaemonSet` vs. `Deployment` — the former is easier to scale out of the box, provided Tyk works best by having just one pod on a node, not winning much from scaling within a node. Of course similar (and probably more robust) result could be achieved with `Deployment`, HPA and anti-affinity rules — it's more difficult to set up properly. So, while I agree it should be configurable, I'd rather leave `DaemonSet` as a default.

PR #11, opened 2019-05-10
More configurable, idiomatic and bugfixed chart

excieve:
Basically addresses #8.

A short list of major changes:
  - No more explicit namespace creation, Helm takes care of that so we just implicitly follow it, allowing more restricted environments (and just common sense with Helm)
  - TLS can be switched on (default) and off with service annotations added for external LB configs (e.g. AWS)
  - Gateway can be configured to be a `DaemonSet` (default) or `Deployment` for more robust scaling options
  - Dashboard ingress now actually works (and can even be served by our ingress controller after initial set up!)
  - Service types are now configurable with gateway being a `LoadBalancer` by default as it should
  - Brought some uniformity in names and labels all around
  - Lots of configurable stuff actually working now
  - Assortment of bug fixes


tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -74,16 +84,26 @@ spec:
             value: "{{ .Values.redis.shardCount }}"
           - name: TYK_GW_STORAGE_HOSTS
             value: "{{ .Values.redis.host }}:{{ .Values.redis.port }}"
+          - name: TYK_GW_STORAGE_PASSWORD
```
lonelycode: Shouldn't this be wrapped in an if statement?

tyk-pro/templates/service-gw.yaml
```diff
@@ -2,20 +2,26 @@ apiVersion: v1
...
   ports:
-  - port: 443
-    targetPort: {{ .Values.gateway.service.port }}
+  - port: {{ .Values.gateway.service.port }}
+    targetPort: 8080
     protocol: TCP
   selector:
     app: gateway-{{ include "tyk-pro.fullname" . }}
     release: {{ .Release.Name }}
-  type: LoadBalancer
```
lonelycode: Do we not want to set this if the service type is set to LB?

PR #35, opened 2019-11-04
Statement regarding headless experimental nature

excieve:


joshblakeley:
Realistically are we ever going to support this officially? Maybe just remove it - it is a nice experiment but unfortunately there is no support for it.

excieve:
I can think of use cases for this, where dashboard is simply not needed — like doing the same thing as nginx ingress controller, but with auth. So I'd rather like this to have at least community-level support at some point, it's just not there yet (and very low priority AFAIK).

PR #65, opened 2020-07-06
Update hybrid gateway ingress resource and bump hybrid chart version

vesko-tyk:
Update Hybrid gateway ingress resource to allow different gateway service ports being used. Atm we're allowing only http, which is causing issue, if you want to use https for example, hence the PR to extend this functionality. Also bump hybrid chart version.


excieve (approved):
:+1:

PR #72, opened 2020-10-30
Headless control api service

Jesse0Michael:
Add add the gateway api port to the gateway values and add an optional service targeting the gateway api, disabled by default.

When the headless gateway has tls disabled, we need to use http instead of https for the Tyk traffic.

---

When I was working with the tyk-operator and the headless tyk gateway deployment, I need to be able to specify a TYK_URL that would reach the CE gateway's control api `TYK_URL=http://gateway-api-svc-tyk-headless.tyk.svc.cluster.local:9696`

~~(LMK if you would rather have the configuration in a different place in the `values_community_edition.yaml` and/or would like a separate service for the control api, instead of adding it to the gateway-svc-tyk-headless... Or if there's a cleaner way to template http vs https Tyk url values)~~
Moved the gateway api service into its own service, disabled by default (in the `values.yaml`), to access the gateway api from the tyk operator


buger:
I would rather keep `values_community_edition.yaml` untouched, and let people explicitly set separate control port if needed.

values_community_edition.yaml
```diff
@@ -29,6 +29,7 @@ gateway:
   hostName: "gateway.tykbeta.com"
   tls: true
   containerPort: 8080
+  controlApiPort: 9696
```
drpebcak, line 32: By default I think this is the same port that the gateway is listening on, so we should set these in such a way that it only overrides if someone intentionally changes it.

values_community_edition.yaml
```diff
@@ -37,6 +38,7 @@ gateway:
   service:
     type: LoadBalancer
     port: 443
+    controlApiPort: 9696
```
drpebcak, line 41: This should be under a different service to avoid exposing the control api.

tyk-headless/templates/configmap-tyk-k8s.yaml
```diff
@@ -13,7 +13,7 @@ data:
       watchNamespaces:
 {{ toYaml .Values.tyk_k8s.watchNamespaces | indent 8 }}
     Tyk:
-      url: "https://localhost:9696"
+      url: "http{{- if .Values.gateway.tls }}s{{- end }}://localhost:9696"
```
drpebcak, line 16: If we are trying to make this configurable, this port should probably not be hardcoded.

drpebcak (changes requested):

tyk-headless/templates/configmap-tyk-k8s.yaml
Jesse0Michael: good point, used `{{ .Values.gateway.api.containerPort }}`

values_community_edition.yaml
Jesse0Michael: separated it out (into the values.yaml) for its own gateway.api.service configuration (disabled by default)

values_community_edition.yaml
Jesse0Michael: 9696 is the default port for the gateway api, moved this to the `values.yaml` file to have the default port explicit (if that's what you meant)

Jesse0Michael:
@buger  I'm not sure if I addressed your concerns, but I made the gateway api explicit in the `values.yaml` leaving the `values_community_edition.yaml` untouched and added a separate service (disabled by default) that the gateway api may be accessed through (but to not expose it on the gateway's load balancer)

Please let me know if i misinterpreted your comment!

values.yaml
```diff
@@ -168,6 +168,12 @@ gateway:
     port: 443
     externalTrafficPolicy: Local
     annotations: {}
+  api:
+    containerPort: 9696
+    service:
+      enabled: false
+      type: NodePort
```
drpebcak, line 175: Probably would be easiest to default to ClusterIP

values.yaml
```diff
@@ -168,6 +168,12 @@ gateway:
     port: 443
     externalTrafficPolicy: Local
     annotations: {}
+  api:
+    containerPort: 9696
```
drpebcak, line 172: This should default to the default for gateway, which is that it listens on the same port that gateway does.

values.yaml
```diff
@@ -168,6 +168,12 @@ gateway:
...
+    service:
+      enabled: false
+      type: NodePort
+      port: 9696
```
drpebcak, line 176: This should default to the default for gateway, which is that it listens on the same port that gateway does.

drpebcak (changes requested):
So to recap, what we need is to not change the current behavior of tyk, but to allow overriding this if we need it to be different.

values.yaml
Jesse0Michael: From what I've experienced, for the headless gateway api the default is 9696. The gateway port that serves traffic through Tyk is 8080.

https://github.com/TykTechnologies/tyk-helm-chart/blob/master/tyk-headless/configs/tyk_mgmt.conf#L3

Are you saying that 9696 is not the default for the gateway api? but something that's manually overridden for the headless configuration?

drpebcak: Interesting.. carry on. This doesn't seem to be the tyk default, but its what was done in the chart I guess.

values.yaml
drpebcak: Based on the other comment, this can be ignored.. Seems like the chart already assumes 9696 in other places.

drpebcak:
@buger thoughts on what is in here now? 9696 appears to be how the gateway was already being configured, so this does not change that.

gernest:
In the operator we deploy two services, one exposing the gateway and another one exposing the control port of the gateway. I  think things are fine as they are by default (exposing the gateway service). What needs to be configurable is the  control service.

What is needed is  to extend gateway section of `values.yaml`  by adding

```yaml
  control:
    enabled: false
    containerPort: 9696
    port: 9696
    type: ClusterIP
```

This will be responsible for configuring control API

and  we can generate the service like this

``` yaml
# tyk-headless/templates/control_service.yaml
{{- if .Values.gateway.control.enabled -}}
apiVersion: v1
kind: Service
metadata:
  name: gateway-control-svc-{{ include "tyk-headless.fullname" . }}
  labels:
    app: gateway-control-svc-{{ include "tyk-headless.fullname" . }}
    chart: {{ include "tyk-headless.chart" . }}
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}
spec:
  type: {{ .Values.gateway.control.type }}
  ports:
  - port: {{ .Values.gateway.control.port }}
    targetPort: {{ .Values.gateway.control.containerPort }}
    protocol: TCP
  selector:
    app: gateway-{{ include "tyk-headless.fullname" . }}
    release: {{ .Release.Name }}
{{- end }}
```

Jesse0Michael:
rebased this PR. Changed the helm values to match Geofrey's suggestions. Removed the non-tls support bit.

tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -63,6 +63,8 @@ spec:
         env:
           - name: TYK_GW_LISTENPORT
             value: "{{ .Values.gateway.containerPort }}"
+          - name: TYK_GW_CONTROLAPIPORT
+            value: "{{ .Values.gateway.control.containerPort }}"
```
gernest, line 67: we also need to add this to the `ports` of the container

```yaml
        ports:
        - containerPort: {{ .Values.gateway.control.containerPort }}
   ```

**Review by gernest — CHANGES_REQUESTED** [review]

File: tyk-headless/templates/deployment-gw-repset.yaml (hunk 7 above)
Comment (Jesse0Michael) [reply tyk-headless/templates/deployment-gw-repset.yaml:67]: added it! (i think)

**Comment by gernest** [comment]
@Jesse0Michael can you please rebase again on master?

Would love to do final review, it looks good so far! I need to install the charts and test before I greenlight this.

**Comment by Jesse0Michael** [comment]
@gernest rebased!

**Comment by hervelemeur** [comment]
@Jesse0Michael FYI there is a typo in this filename: tyk-headless/templates/service-gw-con**rt**ol.yaml

**Comment by Jesse0Michael** [comment]
Whoops! Thanks, @hervelemeur

File: tyk-headless/templates/deployment-gw-repset.yaml [hunk 8]
```diff
@@ -71,6 +71,8 @@ spec:
...
```
excieve, line 75: I think this needs an `if` so that the config is default if the control port is not enabled:
```suggestion
          {{- if .Values.gateway.control.enabled }}
          - name: TYK_GW_CONTROLAPIPORT
            value: "{{ .Values.gateway.control.containerPort }}"
          {{- end }}
```

tyk-headless/templates/service-gw-control.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.gateway.control.enabled -}}
+apiVersion: v1
+kind: Service
+metadata:
```
excieve, line 4: Could you please add the `annotations` too, like on this service?
https://github.com/TykTechnologies/tyk-helm-chart/blob/12213293cf1eaa1f23edfe638167e4181c85c10d/tyk-headless/templates/service-gw.yaml#L10-L15

excieve (changes requested):
Thanks! A couple of things — keeping the default actually disabled through the env vars and adding the annotations to the service would be nice.

tyk-headless/templates/service-gw-control.yaml
Jesse0Michael: added  gateway control annotations

tyk-headless/templates/deployment-gw-repset.yaml
Jesse0Michael: set IFF enabled

excieve (approved):
Looks good to me, thanks a lot!

@gernest any objections to merge this?

gernest (approved):
@Jesse0Michael nice work. I can't wait to use this.

PR #76, opened 2021-01-11
Run containers as non-root by default

vesko-tyk:
Changed the templates to run all Tyk containers as non-root user by default.
Also keeping an option to change the user for the gateway since there might be a use case where this would be needed.
Bumped chart version.

### DIFF
--- a/tyk-headless/configs/tyk_mgmt.conf
+++ b/tyk-headless/configs/tyk_mgmt.conf
@@ -44,7 +44,7 @@
     "allow_master_keys": false,
     "policies": {
         "policy_source": "file",
-        "policy_record_name": "/opt/tyk-gateway/policies/policies.json"
+        "policy_record_name": "/mnt/tyk-gateway/policies/policies.json"
     },
     "hash_keys": true,
     "hash_key_function": "murmur128",
--- a/tyk-headless/configs/tyk_mgmt.conf
+++ b/tyk-headless/configs/tyk_mgmt.conf
@@ -5,14 +5,14 @@
     "node_secret": "$sharedsecret",
     "template_path": "/opt/tyk-gateway/templates",
     "tyk_js_path": "/opt/tyk-gateway/js/tyk.js",
-    "middleware_path": "/opt/tyk-gateway/middleware",
+    "middleware_path": "/mnt/tyk-gateway/middleware",
     "use_db_app_configs": false,
     "db_app_conf_options": {
         "connection_string": "",
         "node_is_segmented": false,
         "tags": []
     },
-    "app_path": "/opt/tyk-gateway/apps/",
+    "app_path": "/mnt/tyk-gateway/apps",
     "storage": {
         "type": "redis",
         "enable_cluster": false,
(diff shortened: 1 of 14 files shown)


tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -85,6 +85,8 @@ spec:
       - name: gateway-{{ .Chart.Name }}
         image: "{{ .Values.gateway.image.repository }}:{{ .Values.gateway.image.tag }}"
         imagePullPolicy: {{ .Values.gateway.image.pullPolicy }}
+        securityContext:
+            runAsNonRoot: true
```
excieve, line 89: This only ensures that k8s won't allow the containers that are attempting to execute as UID 0. We also need to actually execute the container as a non-root user though.

E.g. with this added to the **Pod**-level `securityContext`:
```.yaml
runAsUser: 1000
fsGroup: 2000
```

This applies to all of the deployments in this PR.

For the gateway deployments, you'll also need to make sure that it tried to write the PID file and other files at an accessible location, e.g. like this in the tyk.conf:
```.json
"middleware_path": "/mnt/tyk-gateway/middleware",
"app_path": "/mnt/tyk-gateway/apps",
"pid_file_location": "/mnt/tyk-gateway/tyk.pid"
```

and a special ephemeral mount:
```.yaml
volumeMounts:
  ...
  - name: tyk-scratch
    mountPath: /mnt/tyk-gateway
...
volumes:
  ...
  - name: tyk-scratch
     emptyDir:
       medium: Memory
```

Now you can also add this to the container-level `securityContext`:
```.yaml
allowPrivilegeEscalation: false
privileged: false
readOnlyRootFilesystem: true
capabilities:
  drop:
    - all
```

There is a certain case where you'd want to allow root user execution (when running on a host network), but we don't support this case atm from what I see, so no need to worry about it yet.

Documentation for reference:
https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#podsecuritycontext-v1-core
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#securitycontext-v1-core

excieve (changes requested):
Thanks @vesko-tyk! Please take a look at the comment, it applies to all of the deployments in these charts.

tyk-headless/templates/deployment-gw-repset.yaml
excieve: Re `Memory` ephemeral medium: can be default on-disk, not necessarily in-memory, or even configurable through a value.

vesko-tyk: @excieve Thanks for the review. Got the fixes in my last PR.

excieve (approved):
Thanks @vesko-tyk! Happy to merge if it's tested and working as expected.

vesko-tyk:
I can confirm this is tested and everything is working as expected.

tyk-headless/configs/tyk_mgmt.conf
```diff
@@ -44,7 +44,7 @@
     "allow_master_keys": false,
     "policies": {
         "policy_source": "file",
-        "policy_record_name": "/opt/tyk-gateway/policies/policies.json"    
+        "policy_record_name": "/opt/tyk-gateway/policies/policies.json"
```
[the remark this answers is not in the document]
excieve: @vesko-tyk Makes sense to change this as well since gateway may change it and it'll fail as the root FS is now read-only.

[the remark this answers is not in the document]
vesko-tyk: This is now fixed.

PR #73, opened 2020-11-19
[TD-34][TD-44][TD-39][TD-38] Improve dashboard bootstrap proccess and deprecate tyk-k8s

vesko-tyk:
This is work on issue: https://github.com/TykTechnologies/tyk-helm-chart/issues/12

PR #77, opened 2021-01-21
[TD-44] deprecating tyk-k8s, referring user to tyk-operator

asoorm:


joshblakeley:
Good start here, the helm notes files probably need cleaning up too.

excieve:
I think it depends on #73 in a way as it's more than just documenting the fact.

vesko-tyk:
Inside all charts there is configmaps, ingress resources, secrets, certificates referring to tyk-k8s. All this needs to be cleaned as well. Also in the Readme file we ask people to create namespace named tyk-ingress and use it after it, this would need to be changed as well since we're removing this feature from it.

joshblakeley:
Is there any progress on this @asoorm ?

asoorm:
> Is there any progress on this @asoorm ?

Nop - I thought @vesko-tyk was taking over.

joshblakeley:
> > Is there any progress on this @asoorm ?
> 
> Nop - I thought @vesko-tyk was taking over.

Seem like it is worth catching up together manage expectations on this then otherwise it seems at risk

vesko-tyk:
@joshblakeley @asoorm I believe this PR can be closed now in favour of this one: https://github.com/TykTechnologies/tyk-helm-chart/pull/73/

In PR: https://github.com/TykTechnologies/tyk-helm-chart/pull/73/ I'm addressing the deprecation of tyk-k8s and introducing fully automated bootstrap process.

PR #80, opened 2021-02-27
Istio Ingress Support

joshblakeley:
Tyk can be situated at the edge of a service mesh as the ingress gateway.

These changes make it easy to use our helm charts for this purpose.

All that needs to be done to test is:

Install Istio on a cluster.

To use Tyks gateways as the ingress to your Istio Service Mesh simply change `enableIstioIngress: true` in the values.yaml. Ensure you are using an Istio manifest which disables the default Istio Ingress gateway.

There will be a docs guide shortly that goes through the whole process based on these changes and also includes using the operator.

Also have added some sample applications to reference from docs and guides.

Also change the gateway config so we are able to use tyk sync and the operator by default with the gateway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,3 +107,7 @@ If you are running an older chart that does not have this value, then you can di
 NB: tyk-k8s has been deprecated. For reference, old documentation may be found here: [Tyk K8s](https://github.com/TykTechnologies/tyk-k8s)
 
 For further detail on how to configure Tyk as an Ingress Gateway, or how to manage APIs using the Kubernetes API, please refer to our [Tyk Operator documentation](https://github.com/TykTechnologies/tyk-operator/).
+
+## Istio Service Mesh Ingress gateway
+
+To use Tyks gateways as the ingress to your Istio Service Mesh simply change `enableIstioIngress: true` in the values.yaml. Ensure you are using an Istio manifest which disables the default Istio Ingress gateway.
--- a/samples/httpbin.yaml
+++ b/samples/httpbin.yaml
@@ -35,23 +35,23 @@ spec:
         name: httpbin
         ports:
         - containerPort: 80
----
-apiVersion: extensions/v1beta1
-kind: Ingress
-metadata:
-  name: httpbin-ingress
-  annotations:
-    kubernetes.io/ingress.class: tyk
-spec:
-  rules:
-  - host: www.tyk-test.com
-    http:
-      paths:
-      - path: /httpbin
-        backend:
-          serviceName: httpbin
-          servicePort: 80
-      - path: /httpbin2
-        backend:
-          serviceName: httpbin
-          servicePort: 8000
+# ---
+# apiVersion: networking.k8s.io/v1
+# kind: Ingress
+# metadata:
+#   name: httpbin-ingress
+#   annotations:
+#     kubernetes.io/ingress.class: tyk
+# spec:
+#   rules:
+#   - host: www.tyk-test.com
+#     http:
+#       paths:
+#       - path: /httpbin
+#         backend:
+#           serviceName: httpbin
+#           servicePort: 80
+#       - path: /httpbin2
+#         backend:
+#           serviceName: httpbin
+#           servicePort: 8000
--- a/samples/httpbin.yaml
+++ b/samples/httpbin.yaml
@@ -1,0 +1,57 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: httpbin
+  labels:
+    app: httpbin
+spec:
+  ports:
+  - name: http
+    port: 8000
+    targetPort: 80
+  selector:
+    app: httpbin
+---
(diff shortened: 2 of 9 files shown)


joshblakeley:
Want to try this out? Docs are available here: https://github.com/TykTechnologies/tyk-docs/pull/1567/files

README.md
```diff
@@ -107,3 +107,7 @@ If you are running an older chart that does not have this value, then you can di
 NB: tyk-k8s has been deprecated. For reference, old documentation may be found here: [Tyk K8s](https://github.com/TykTechnologies/tyk-k8s)
 
 For further detail on how to configure Tyk as an Ingress Gateway, or how to manage APIs using the Kubernetes API, please refer to our [Tyk Operator documentation](https://github.com/TykTechnologies/tyk-operator/).
+
+## Istio Service Mesh Ingress gateway
+
+To use Tyks gateways as the ingress to your Istio Service Mesh simply change `enableIstioIngress: true` in the values.yaml. Ensure you are using an Istio manifest which disables the default Istio Ingress gateway.
```
vesko-tyk, line 113: Shouldn't we include link to our docs where people can see examples of how they can use Tyk as Ingress Gateway for Istio on k8s?

samples/httpbin.yaml
```diff
@@ -0,0 +1,57 @@
...
+        version: v1
+    spec:
+      containers:
+      - image: docker.io/kennethreitz/httpbin
+        imagePullPolicy: IfNotPresent
+        name: httpbin
+        ports:
+        - containerPort: 80
+---
+apiVersion: extensions/v1beta1
```
vesko-tyk, line 39: `extensions/v1beta1` ingress is deprecated in v1.14+ and will be unavailable in v1.22+. We should use `networking.k8s.io/v1` ingress instead.

tyk-pro/templates/deployment-gw-repset.yaml
```diff
@@ -30,6 +30,13 @@ spec:
       release: {{ .Release.Name }}
   template:
     metadata:
+      {{- if .Values.enableIstioIngress }}
+      annotations:
+        traffic.sidecar.istio.io/excludeInboundPorts: "{{ .Values.gateway.containerPort}}"
+        traffic.sidecar.istio.io/includeInboundPorts: ""
+        traffic.sidecar.istio.io/excludeOutboundPorts: "{{ .Values.dash.containerPort }}"
+        traffic.sidecar.istio.io/includeOutboundPorts: ""
```
vesko-tyk, line 38: Things are a bit unclear for me here: By definition `excludeInboundPorts` is: `A comma separated list of inbound ports to be excluded from redirection to Envoy. Only applies when all inbound traffic (i.e. '*') is being redirected.`, but then we set all inbound traffic to be disabled by giving it an empty list eg. `includeInboundPorts: ""`, so shouldn't it the wildcard character '*' be used here to configure redirection for all inbound ports?

Same applies for outbound ports. Here is docs for reference: https://istio.io/latest/docs/reference/config/annotations/

tyk-pro/templates/deployment-dash.yaml
```diff
@@ -25,6 +25,11 @@ spec:
       labels:
         app: dashboard-{{ include "tyk-pro.fullname" . }}
         release: {{ .Release.Name }}
+      {{- if .Values.enableIstioIngress }}
+      annotations:
+        traffic.sidecar.istio.io/excludeInboundPorts: "{{ .Values.dash.containerPort }}"
+        traffic.sidecar.istio.io/includeInboundPorts: ""
```
vesko-tyk, line 31: Please see comment in `deployment-gw-repset.yaml` file regarding `InboundPorts`. Same question applies here as well.

vesko-tyk (changes requested):
Great work @joshblakeley. 

Just got few comments, if you can review them that would be great.

Thanks!

README.md
joshblakeley: Once those docs are published then yes i'll come back do it - nothing we can do yet though

samples/httpbin.yaml
joshblakeley: Sure thing - i'm going to comment the ingress def out for now too so its optional if someone wants to try it out and they have the operator for instance

tyk-pro/templates/deployment-gw-repset.yaml
joshblakeley: It a bug in the istio implementation which means the defined excludeInboundPorts wont work unless you also provide the other annotation even if it's empty and does nothing.

joshblakeley: I'll have a play and see if wildcard affects things - although i'm comfortable leaving the possibility of all outward facing ports being available because that makes opening up TCP/gRPC apis a lot simpler

vesko-tyk (approved):
LGTM

PR #81, opened 2021-03-02
Hybrid improvements

joshblakeley:
Some improvements to docs and useability aimed at improving the get started journey for Tyk Cloud users.

 - Move the RPC key and API key fields to the gateway RPC section
 - insecure skip verify is default
 - remove bindtoslugs
 - move helm install command up and style like the others
 - Change lb service to nodeport to match other charts
 - add some helpful comments and an example config so that classic cloud users can keep using these charts

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,6 @@ To install, *first modify `values_hybrid.yaml` file as follows:*
 > If you are using hybrid gateway with the Tyk Classic Cloud use the rpc settings block commented out in the values yaml
 
 
-
 ## Installing TIB
 
 TIB is not necessary to install for Pro installations and it's functionality is included in the Tyk Dashboard API Manager.
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ To install, *first modify `values_hybrid.yaml` file as follows:*
 > If you are using hybrid gateway with the Tyk Classic Cloud use the rpc settings block commented out in the values yaml
 
 
+
 ## Installing TIB
 
 TIB is not necessary to install for Pro installations and it's functionality is included in the Tyk Dashboard API Manager.
--- a/README.md
+++ b/README.md
@@ -49,6 +49,18 @@ To install, *first modify the `values.yaml` file to add your license*:
 
 Follow the instructions in the Notes that follow the installation to find your Tyk login credentials.
 
+## Install Tyk Hybrid Gateways (This can be used either for Hybrid Gateways connected to Tyk Cloud or MDCB Hybrid Gateways)
+
+To install, *first modify `values_hybrid.yaml` file as follows:*
+1. Add your dashboard users organisation ID in `gateway.rpc.orgID` value
+2. Add your dashboard users API key in `gateway.rpc.apiKey` value 
+3. Add your connection string to allow the Hybrid gateway to connect to your control plane in `gateway.rpc.connString`. On the Tyk Cloud Console find this value in the endpoints panel for your control plane deployment.
+
+	`helm install tyk-hybrid -f ./values_hybrid.yaml ./tyk-hybrid -n tyk`
+
+> If you are using hybrid gateway with the Tyk Classic Cloud use the rpc settings block commented out in the values yaml
+
+
 ## Installing TIB
 
 TIB is not necessary to install for Pro installations and it's functionality is included in the Tyk Dashboard API Manager.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -49,6 +49,14 @@ To install, *first modify the `values.yaml` file to add your license*:
 
 Follow the instructions in the Notes that follow the installation to find your Tyk login credentials.
 
+## Install Tyk Hybrid Gateways (This can be used either for Hybrid Gateways connected to Tyk Cloud or MDCB Hybrid Gateways)
+
+To install, *first modify `values_hybrid.yaml` file as follows:*
+1. Add your dashboard users organisation ID in `gateway.rpc.orgID` value
+2. Add your dashboard users API key in `gateway.rpc.apiKey` value
```
excieve, line 56: This should mention setting `gateway.rpc.connString` too, given that it's now variable.

tyk-hybrid/templates/deployment-gw-repset.yaml
```diff
@@ -78,17 +78,19 @@ spec:
           - name: TYK_GW_HTTPSERVEROPTIONS_USESSL
             value: "{{ .Values.gateway.tls }}"
           - name: TYK_GW_SLAVEOPTIONS_RPCKEY
-            value: "{{.Values.tyk_k8s.org_id }}"
+            value: "{{.Values.rpc.rpcKey }}"
```
excieve, line 81: Since the variable has changed, something like this should keep compat:
```suggestion
            value: "{{ default .Values.tyk_k8s.org_id .Values.rpc.rpcKey }}"
```

tyk-hybrid/templates/deployment-gw-repset.yaml
```diff
@@ -78,17 +78,19 @@ spec:
...
           - name: TYK_GW_SLAVEOPTIONS_APIKEY
-            value: "{{ .Values.tyk_k8s.dash_key }}"
+            value: "{{ .Values.rpc.apiKey }}"
```
excieve, line 83:
```suggestion
            value: "{{ default .Values.tyk_k8s.dash_key .Values.rpc.apiKey }}"
```

values_hybrid.yaml
```diff
@@ -22,27 +20,40 @@ redis:
...
+  tags: ""
   hostName: ""
   tls: true
   containerPort: 8080
-  # For MDCB setups change to connection string for your MDCB instance
+  # For on premise MDCB setups sipmly change the connection string to point to your MDCB instance
   rpc:
-    connString: "hybrid.cloud.tyk.io:9091"
+    # On the Tyk Cloud Console find this value in the endpoints for your control plane deployment
+    connString: "MDCB-INGRESS:443"
```
excieve, line 37: What's the reasoning to change this default to a dummy value?

values_hybrid.yaml
```diff
@@ -22,27 +20,40 @@ redis:
...
   # Set replicaCount for Deployments
   replicaCount: 2
-  tags: "ingress"
+  # if enableSharding set to true then you must define a tag to load APIs to these gateways i.e. "ingress"
+  tags: ""
   hostName: ""
   tls: true
   containerPort: 8080
-  # For MDCB setups change to connection string for your MDCB instance
+  # For on premise MDCB setups sipmly change the connection string to point to your MDCB instance
```
excieve, line 34:
```suggestion
  # For on premise MDCB setups simply change the connection string to point to your MDCB instance
```

excieve (changes requested):
Looks good, thanks @joshblakeley!

I'd prefer if we could have some compatibility with the previous RPC vars though, so that we don't make people's lives worse if they're already using this chart.

Please bump the chart version too.

values_hybrid.yaml
joshblakeley: New cloud users need to take precedence and every single person i have interacted with so far assumes that `hybrid.cloud.tyk.io:9091` will magically work for them with Ara. 

I would rather they have to read the comment and think - if they need the classic cloud ingress i would prefer they use the whole commented block underneath with that configuration

joshblakeley: I can just make it blank for now

joshblakeley:
@excieve i have a few more PRs open and coming through the pipeline today - shall I hold off bumping the chart version until theyre all in?

excieve (approved):
Thanks @joshblakeley! Regarding the version bump — I'd do it on each PR as we don't have a proper chart repository and tagging yet (that's coming) so exposed version is currently essentially the same as the state of the code in the git repo.

joshblakeley:
Thanks @excieve have bumped the hybrid chart version. I'll leave it to your team to merge please.

excieve (approved):

PR #83, opened 2021-03-06
OSS Deployment improvements

joshblakeley:
There are some weird things with the headless helm chart i'm trying to address.

1. Disable sharding and tags - with the OSS gateway the default behaviour is based around 1 GW so lets not introduce that complexity. Having it enabled is an artifact from the deprecated service mesh piece.
2. For some reason the chart assumes a mongo is deployed even though we say it is not necessary in the readme - this means we are harming gateway performance by producing analytics and either sending to Redis where it is purged or we send to a Mongo instance which relies on a Tyk Dashboard to visualise. Instead we should look at how to accommodate open source options for defaults in another PR. Suggestions welcome!

### DIFF
--- a/values_community_edition.yaml
+++ b/values_community_edition.yaml
@@ -6,7 +6,7 @@ nameOverride: ""
 fullnameOverride: "tyk-headless"
 
 # Only set this to false if you are not planning on using the sidecar injector
-enableSharding: true
+enableSharding: false
 
 secrets:
   APISecret: "CHANGEME"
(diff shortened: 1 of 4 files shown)


values_community_edition.yaml
```diff
@@ -6,42 +6,42 @@ nameOverride: ""
 fullnameOverride: "tyk-headless"
 
 # Only set this to false if you are not planning on using the sidecar injector
-enableSharding: true
+enableSharding: false
```
vesko-tyk, line 9: Could you please update the comment here or remove it altogether?

vesko-tyk (changes requested):
Thanks @joshblakeley. Got one small comment and also, if you can bump chart version that would be amazing.

joshblakeley:
Thanks for review @vesko-tyk have bumped chart version and removed the confusing sharding value and comment since theyre not relevant for now until we have a way of users to manage clustered gateway deployments in k8s via the operator for example

vesko-tyk (approved):
LGTM

PR #86, opened 2021-03-08
Fix Headless deployment

vesko-tyk:


tyk-headless/templates/deployment-pmp.yaml
```diff
@@ -47,14 +48,22 @@ spec:
         env:
           - name: REDIGOCLUSTER_SHARDCOUNT
             value: "{{ .Values.redis.shardCount }}"
+            {{- if .Values.mongo.enabled }}
```
joshblakeley, line 51: Wrap all of these in a single if statement since they all follow the same behaviour

joshblakeley (changes requested):
Once small comment but otherwise great thanks @vesko-tyk !

tyk-headless/templates/deployment-pmp.yaml
vesko-tyk: done

joshblakeley (approved):
💯

PR #87, opened 2021-03-08
[TD-43] Restructure charts to make use of the default values

vesko-tyk:


joshblakeley:
Will need to update all tyk docs references to the helm chart installations  - otherwise looks good thanks @vesko-tyk

excieve (approved):
Looks good, thanks! It's probably better to instruct to copy the `values.yaml` or create your own rather than modify one directly in the repo though.

PR #91, opened 2021-03-15
[TD-37] Roll using rand method instead of template hash

joshblakeley:
The previous roll method caused recursion errors because of self referencing templates.

Instead to self roll a dep you can use the rand method detailed in the helm docs here: https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments

We should also document the behaviour for helm upgrade that you get because of this since before someone would need the deprecated --recreate-pods flag to roll their pods

### DIFF
--- a/tyk-headless/templates/deployment-gw-repset.yaml
+++ b/tyk-headless/templates/deployment-gw-repset.yaml
@@ -34,7 +34,7 @@ spec:
         app: gateway-{{ include "tyk-headless.fullname" . }}
         release: {{ .Release.Name }}
       annotations:
-        checksum/config: {{ include (print $.Template.BasePath "/deployment-gw-repset.yaml") . | sha256sum }}
+        rollme: {{ randAlphaNum 5 | quote }}
     spec:
 {{- if .Values.gateway.nodeSelector }}
       nodeSelector:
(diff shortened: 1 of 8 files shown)


drpebcak:
What resources are changing without causing these pods to get recreated? The configmaps? If that's the case, why not just reference a checksum of the configmap instead? Would be better than forcing re-rolls always, when it might not be needed.

joshblakeley:
The original intention seems to have just been to roll when the deployment itself was modified not any other file. I've jsut followed the recommended approach for this from the Helm docs but would welcome input from @excieve and @vesko-tyk

drpebcak:
For a deployment the default behavior of kubernetes would already cause the pods to re-roll if anything inside the deployment object is changed.

excieve (changes requested):
The idea was to hash the configmaps, not the deployments. It seems the includes in the annotations are referencing the deployment yamls, which is of course wrong.

Rolling each time is not great as noted in this thread. Would be great if the PR instead referenced correct templates.

I will also reopen the internal ticket that tracked the original change.

joshblakeley:
Closing so this can be addressed by the devops team

PR #93, opened 2021-03-15
tyk-headless: Mount apps,middleware and policies directories

gernest:
For tyk-headless deploying the chart and trying to create new api results in
 the following error

 ```
Failed to create file! - open /mnt/tyk-gateway/apps/YmRkL3dvcmthcm91bmQtYXBp.json: no such file or directory
 ```

The gateway assumes `/mnt/tyk-gateway/apps` is a directory that already exists,
however we only mount `/mnt/tyk-gateway` which means we need to create `apps`
sub directory for thw gateway to work

 This PR ensures that `apps`, `middleware` and `policies` directories are
 properly mounted making the gateway happy when creating/updating and deleting
 api's and policies

With this patch, the error above will not happen and the gateway will work as expected

### DIFF
--- a/tyk-headless/templates/deployment-gw-repset.yaml
+++ b/tyk-headless/templates/deployment-gw-repset.yaml
@@ -104,6 +104,12 @@ spec:
             mountPath: /etc/certs
           - name: tyk-scratch
             mountPath: /mnt/tyk-gateway
+          - name: tyk-scratch
+            mountPath: /mnt/tyk-gateway/apps
+          - name: tyk-scratch
+            mountPath: /mnt/tyk-gateway/middleware
+          - name: tyk-scratch
+            mountPath: /mnt/tyk-gateway/policies
         livenessProbe:
           httpGet:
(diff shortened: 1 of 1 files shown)


excieve (commented):
Thanks @gernest!

I think there's a better way to do this by setting up an [init container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) that basically runs `mkdir -p` for all the required paths once the pod is created before the app container runs.

Sub-directory mounts (e.g. `/mnt/tyk-gateway/apps`) under another mount (`/mnt/tyk-gateway`) can be conflicting. Also this results in allocating 4 separate volumes that shouldn't be really needed.

I can imagine different volumes being a use case if a user wants to provision the apps/policies/middleware from the files, but it would need more configurability (e.g. Helm values that point to user-provided configmaps) and potentially other considerations (such as an init container that copies the provisioned files to avoid conflicts). So this is perhaps for another PR.

gernest:
@excieve thanks, I opened a separate PR which uses initContainers to setup directories

Closing this in favour of #98

PR #84, opened 2021-03-08
[TD-37] Automatically roll deployments

vesko-tyk:

PR #95, opened 2021-03-17
[TD-37] Checksum configmaps

gernest:
Fixes #84

Fixes checksum/config annotation to use correct config maps instead of self
referencing deployment (That was leading to recursion)

### DIFF
--- a/tyk-headless/templates/deployment-gw-repset.yaml
+++ b/tyk-headless/templates/deployment-gw-repset.yaml
@@ -34,7 +34,7 @@ spec:
         app: gateway-{{ include "tyk-headless.fullname" . }}
         release: {{ .Release.Name }}
       annotations:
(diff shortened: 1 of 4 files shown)


jack-of-AllTrades:
Hey @gernest! i think you're missing a few deployments for tyk-headless and tyk-pro.

gernest:
@jack-of-AllTrades which ones ?

gernest:
Just to clarify, this PR only fixes the issue where we were checksumming wrong files. Covering checksumming all configmaps is beyond the scope of this PR and can be done separately.

jack-of-AllTrades:
i think the checksum is referring to itself for all deployment files. I'll submit a new PR for rest of the files.

excieve (approved):
Thanks @gernest!

PR #96, opened 2021-03-17
remove mongo and redis install instructions

gernest:
The instructions are incorrect, mongo and redis manifests are no longer part of the project.
Users should figure out how they install redis and mongodb on their own.


excieve (commented):
@gernest Sorry, why do you think they're not a part?

Simple manifests are still available here (and were actually added recently):
https://github.com/TykTechnologies/tyk-helm-chart/tree/master/deploy/dependencies

Instructions for more production-like Bitnami charts are also useful I believe.

gernest:
oops! for some reason that directory wasn't showing up on my vscode

PR #97, opened 2021-03-17
TD-37 - Pods dont automatically pick up config changes

jack-of-AllTrades:
This one along with #95 should fix TD-37 ticket


excieve (approved):
Thanks @jack-of-AllTrades!

PR #98, opened 2021-03-17
tyk-headless: Setup directories needed for tyk to be operational

gernest:
Based on the solution proposed here https://github.com/TykTechnologies/tyk-helm-chart/pull/93#pullrequestreview-614090651

For tyk-headless deploying the chart and trying to create new api results in
the following error

 ```
Failed to create file! - open /mnt/tyk-gateway/apps/YmRkL3dvcmthcm91bmQtYXBp.json: no such file or directory
 ```

The gateway assumes `/mnt/tyk-gateway/apps` is a directory that already exists,
however we only mount `/mnt/tyk-gateway` which means we need to create `apps`
sub directory for thw gateway to work

 This PR ensures that `apps`, `middleware` and `policies` directories are
 properly created making the gateway happy when creating/updating and deleting
 api's and policies

This used init container to setup the directories

### DIFF
--- a/tyk-headless/templates/deployment-gw-repset.yaml
+++ b/tyk-headless/templates/deployment-gw-repset.yaml
@@ -48,6 +48,14 @@ spec:
       affinity:
 {{ toYaml .Values.gateway.affinity | indent 8 }}
 {{- end }}
+      initContainers: 
+      - name: "setup-directories"
+        image: busybox
+        command: ['sh','-c','mkdir -p apps && mkdir -p middlewares && mkdir -p policies']
+        workingDir: /mnt/tyk-gateway
+        volumeMounts:
+          - name: tyk-scratch
+            mountPath: /mnt/tyk-gateway        
       containers:
       - name: gateway-{{ .Chart.Name }}
         image: "{{ .Values.gateway.image.repository }}:{{ .Values.gateway.image.tag }}"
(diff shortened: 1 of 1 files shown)


tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -48,6 +48,14 @@ spec:
       affinity:
 {{ toYaml .Values.gateway.affinity | indent 8 }}
 {{- end }}
+      initContainers: 
+      - name: "setup-directories"
+        image: busybox
```
excieve, line 53: Always best to specify a version of an image to avoid unexpected behaviour in the future:
```suggestion
        image: busybox:1.32
```

tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -48,6 +48,14 @@ spec:
...
+        command: ['sh','-c','mkdir -p apps && mkdir -p middlewares && mkdir -p policies']
```
excieve, line 54: This can be simplified:
```suggestion
        command: ['sh','-c','mkdir -p apps middleware policies']
```

Also seems like the config expects `/mnt/tyk-gateway/middleware` (not `middlewares`):
https://github.com/TykTechnologies/tyk-helm-chart/blob/fdff092f17291f16e4f3823c256ac2c91418e8c3/tyk-headless/configs/tyk_mgmt.conf#L8

excieve (changes requested):
Very nice, thanks @gernest! Just a couple of minor things.

gernest:
@excieve thanks, I applied the changes PTAL

excieve (approved):
Looks good, thanks!

PR #100, opened 2021-03-22
Add github actions to push Tyk-Pro, Tyk-Hybrid and Tyk-Headless chart…

gwotg:
Adding GH Actions to push all 3 charts to CloudSmith's helm repo.


excieve (commented):
Thanks @gwotg! If I understand this right (correct me if I'm wrong), this only pushes, but doesn't build the archives. Shouldn't we build in the same workflows too?

Also I think we should only push on release. There's no process for this currently, but I can guess that it will depend on the git tags, so perhaps let the actions trigger on that rather than master/main/other branch?

A potential tag could look like `v1.0.0-tyk-pro` or `v1.0.0-tyk-headless`, etc. Then the action would know whether to release the `tyk-pro`, `tyk-headless` or `tyk-hybrid` chart (although at this point I'm not entirely sure why do we need three instead of one, but that's a different discussion).

And based on this scheme, can we have just one workflow that knows what to push from the tag name instead of three?

Happy to discuss more regarding releases/versioning.

alephnull (changes requested):
Since the steps are identical, implementing it as a matrix step over hybrid, pro and headless would collapse this down to one file.
An implementation using a matrix step to achieve a similar usecase is at https://github.com/TykTechnologies/tyk/blob/release-3-lts/.github/workflows/sync-automation.yml

gwotg:
Thanks for jumping in folks.

@excieve You're right - I had a few questions and you seem to have addressed all of them right away - I have no idea whatsoever on how to version them. I also need to know if we're up and about to use runs within the action or on scripts - the run bit that installs helm and gen the package - scripts would allow me to not repeat myself even more. I can totally set it conditional to the tag if this is what we're planning to go with.

@alephnull Regarding the loop, I'm making the run conditional to where the changes were carried out - do you see any options to it? This is a multi-repo where we need to track multiple "builds" on the same repo, not all to be updated at once.

.github/workflows/tyk-pro.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Push
+        id: push
+        uses: cloudsmith-io/action@master
+        with:
+          api-key: ${{ secrets.CLOUDSMITH_API_KEY }}
+          command: 'push'
+          format: 'helm'
+          owner: 'tyk'
+          repo: 'tyk-pro'
+          version: '0.0.0'
```
alephnull, line 46: Shouldn't this version correspond to the tag that was pushed?

alephnull (commented):
Does it make sense to allow running [these actions manually](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch)?

.github/workflows/tyk-pro.yaml
gwotg: Hi @alephnull, thanks for chiming in! 

You're right, just didn't want to wait for a decision on how exactly we'll manage that to prevent a successful run. That means a release process to be created/maintained. 

Do you have an example mechanism I could base this one of? Not the actions per se, I've read your previous example and understand the implementation/triggers/variables - I mean the whole remaining organization, like maintainers, cadence, etc.

alephnull:
Here's a thoroughly untested implementation of Artem's proposal for just tyk pro:

```diff
diff --git a/.github/workflows/tyk-pro.yaml b/.github/workflows/tyk-pro.yaml
index ff979a7..33edbe1 100644
--- a/.github/workflows/tyk-pro.yaml
+++ b/.github/workflows/tyk-pro.yaml
@@ -1,14 +1,18 @@
 name: Tyk Pro Push Helm
-on: push
-jobs:
+
+on:
   push:
-    name: Tyk Pro Push CS
-    runs-on: ubuntu-latest
+    tag: 'v*-pro'
     branches:
       - main
       - release-**
     paths:
       - 'tyk-pro/**'
+
+jobs:
+  push:
+    name: Tyk Pro Push CS
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
         uses: actions/checkout@v2
@@ -28,10 +32,11 @@ jobs:
 
       - name: Pack helm artifact
         run: |
-          output=$(helm package ${CHART} --version ${CHART_VERSION})
+          tag=${GITHUB_REF##*/}
+          output=$(helm package ${CHART} --version ${tag})
           echo $output
+          echo "version=$tag" >> $GITHUB_ENV
         env:
-          CHART_VERSION: '0.0.0'
           CHART: tyk-pro
 
       - name: Push
@@ -43,5 +48,5 @@ jobs:
           format: 'helm'
           owner: 'tyk'
           repo: 'tyk-pro'
-          version: '0.0.0'
-          file: './tyk-pro-0.0.0.tgz'
+          version: ${{ env.version }}
+          file: './tyk-pro-${{ env.version }}.tgz'
```

This implementation based on tags is somewhat opposed to,
> This is a multi-repo where we need to track multiple "builds" on the same repo, not all to be updated at once.

Since tags are global and if the charts are independently versioned, we have to resort to ugly hacks like tacking `-pro` to the end of tags. However, if they are split into different repos, the patch above should work without having to use a path trigger or have an ugly -pro suffix in user-visible versions.

.github/workflows/tyk-pro.yaml
alephnull: Seems to me that you're in a position to define the process :)

Please add helm to https://tyktech.atlassian.net/wiki/spaces/EN/pages/449708061/Release+Engineering

.github/workflows/tyk-headless.yaml
```diff
@@ -0,0 +1,52 @@
+name: Tyk Headless Push Helm
+on: push
+jobs:
+  push:
+    name: Tyk Headless Push CS
+    runs-on: ubuntu-latest
+    branches:
+      - main
```
abhishiekgowtham, line 8: i think this should refer to master instead of main. Similar thing for rest of workflow files

.github/workflows/tyk-headless.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - 'tyk-headless/**'
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set env
+        run: |
+          VER=$(sed 's/[^0-9.]//g' <<< GITHUB_REF##*/)
+          echo "RELEASE_VERSION=${VER}" >> $GITHUB_ENV
+
+      - name: Install Helm, create dirs
```
abhishiekgowtham, line 21: Isn't it easier to use [this](https://github.com/marketplace/actions/helm-tool-installer) action instead?

gwotg: Could be a ggood addition, gotta try it

.github/workflows/tyk-headless.yaml
gwotg: Once on master we'll be running it on every commit.

.github/workflows/tyk-headless.yaml
excieve: @gwotg Would be nice to use it like in the other workflows, e.g. here:
https://github.com/TykTechnologies/tyk-helm-chart/blob/5923a801bc407fedea34f0b5ce27d4d9a65b933c/.github/workflows/smoke-test.yml#L100-L103

excieve (changes requested):
Thanks! I think we shouldn't trigger the archive on master/main push, but only on the version tags and the `release-` branch (just in case, although not sure it will be used much).

Also, any way we could do just one workflow as these are essentially a copy paste? Could be in a later improvement to get this fleshed out first.

typekpb:
anyone willing to adress the issues raised? As it's almost 1 month since the provided feedback

abhishiekgowtham:
Hey @typekpb! We're working on this, and will be released as part of v1.0.0 tag.

PR #101, opened 2021-03-23
injects portal path as config env variable into dashboard.

sedkis:
The Tyk Developer Portal `path` currently being read from the `tyk_analytics.conf`. This PR changes this in order to inject the portal path from the `values.yaml` into the Dashboard as an environment variable.

Also, I moved the `portal.ingress.path` to `portal.path` because the value applies to the Portal globally, even outside of any Ingress objects created.  Furthermore, there is a `portal.ingress.enabled` section which will be confusing if that is set to `false`, yet we are still injecting the `path` into an environment variable.


excieve (approved):
Makes sense, thanks @sedkis!

PR #102, opened 2021-03-24
tyk-pro: ensure checksum/config for pump deployment is set

gernest:
This change make sure that checksum/config is set even when
`.Values.pump.annotations` is not set

To reproduce the issue, just comment  `annotations` line in `pump` section of `values.yaml` this will create the pod without the checksum annotation.


excieve (approved):
Good catch, thanks @gernest!

PR #103, opened 2021-03-25
tyk-pro: enable installing tyk in a different namespace

### LINKED ISSUE #92: Remove dependency on namespace "tyk"
If you try to deploy `tyk-pro` outside of the `tyk` namespace, it gives you many errors related to namespace "tyk". 

You should be able to install it anywhere.  If I have multiple namespaces in my team, this is problematic.

Can we please remove the hard-coded "tyk" namespace

gernest:
Closes #92 

 This PR enables tyk-pro to be installed in any  namespace  specified on helm install.

Currently you can only  install tyk-pro chart in `tyk` namespace.


excieve (approved):
Very nice, thanks @gernest!

Will merge it, but let's pick up the script templating discussion in #105.

PR #105, opened 2021-03-26
[TD-219] - Tyk Helm: implement full Tyk Pro bootstrap

abhishiekgowtham:
Use apis to bootstrap dashboard and developer portal


gernest:
Interesting, I also had a different approach to the bootstrap script regarding using correct namespace and stuff on #103
Basically I treat the script  as a template so you have access to all helm template benefits .

Instead of `NAMESPACE=$TYK_POD_NAMESPACE` you can do `NAMESPACE={{.Release.Namespace}}`

abhishiekgowtham:
Oh nice. I tried to use them as env vars in the script. But i think changes to namespace and other things are out of scope for this PR. I think maybe once this gets merged then can create an internal ticket to work on the improvements/suggestions.

tyk-pro/templates/NOTES.txt
```diff
@@ -2,8 +2,8 @@
 
 For the URL: (kubectl get nodes --selector=kubernetes.io/role!=master -o jsonpath='{.items[0].status.addresses[?(@.type=="ExternalIP")].address}'), if you're using Minikube then: (minikube ip) would be sufficient"
 For the port: (kubectl get --namespace tyk -o jsonpath="{.spec.ports[0].nodePort}" services dashboard-svc-tyk-pro)"
-For username: (kubectl get secret --namespace tyk tyk-login-details -o jsonpath="{.data.TYK_USER}" | base64 --decode)
-For password: (kubectl get secret --namespace tyk tyk-login-details -o jsonpath="{.data.TYK_PASS}" | base64 --decode)
+For username: "{{ .Values.dash.adminUser.email }}"
+For password: (kubectl get secret --namespace tyk {{ .Release.Name }}-login-details -o jsonpath="{.data.TYK_PASS}" | base64 --decode)
```
excieve, line 6:
```suggestion
For password: (kubectl get secret --namespace {{ .Release.Namespace }} {{ .Release.Name }}-login-details -o jsonpath="{.data.TYK_PASS}" | base64 --decode)
```

tyk-pro/templates/secret-dashboard.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: {{ .Release.Name }}-login-details
+  labels:
+    app: login-details-{{ include "tyk-pro.fullname" . }}
+    chart: {{ include "tyk-pro.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+type: Opaque
+data:
+  TYK_PASS: {{ randAlphaNum 10 | b64enc | quote }}
```
excieve, line 12: I think it would be useful for the user to set this optionally through a value too, while defaulting to a random.

tyk-pro/scripts/bootstrap_k8s.sh
```diff
@@ -1,28 +1,176 @@
...
 
 set -e
 
-apk --no-cache add curl
+apk --no-cache add curl jq
 curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
 install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
 
-NAMESPACE=tyk
+NAMESPACE=$TYK_POD_NAMESPACE
```
excieve, line 9: This `NAMESPACE` variable isn't really needed, `TYK_POD_NAMESPACE` can be used directly.

tyk-pro/scripts/bootstrap_k8s.sh
```diff
@@ -1,28 +1,176 @@
...
+      ;;
+  esac
+}
+
+if_present_echo(){
+  [[ -z "$1" ]] && return $2 || echo "$1"
+}
+
+create_organisation() {
+  ORGDATA=$(curl --silent --header "admin-auth: $TYK_ADMIN_SECRET" --header "Content-Type:application/json" --data '{"owner_name": "'"$TYK_ORG_NAME"'", "cname_enabled": true, "cname": "'$TYK_ORG_CNAME'"}' $DASHBOARD_HOSTNAME/admin/organisations 2>&1)
```
excieve, line 82: `DASHBOARD_HOSTNAME` variable doesn't appear to include the protocol as seen above:
```
DASHBOARD_HOSTNAME="$TYK_DASHBOARD_SVC.$TYK_POD_NAMESPACE.svc.cluster.local:$TYK_DB_LISTENPORT"
```

Does it work correctly? Same question on the other usages.

tyk-pro/scripts/bootstrap_k8s.sh
```diff
@@ -1,28 +1,176 @@
...
-
-kubectl create secret -n ${NAMESPACE} generic tyk-login-details \
-  --from-literal "TYK_USER=${TYK_USER}" \
-  --from-literal "TYK_PASS=${TYK_PASS}"
+  --from-literal "TYK_AUTH=${USER_AUTH_CODE}" \
+  --from-literal "TYK_ORG=${ORGID}" \
+  --from-literal "TYK_MODE=pro" \
+  --from-literal "TYK_URL=http://${DASHBOARD_HOSTNAME}"
+
+if [ "$BOOTSTRAP_PORTAL" = "true" ]; then
```
excieve, line 113: I don't think there's a need to put the function definitions under the conditionals.

Also, would be good to refactor this a little so that all the necessary bootstrapping happens in the `main()` function, where the portal parts are under the conditional.

tyk-pro/scripts/bootstrap_k8s.sh
```diff
@@ -1,28 +1,176 @@
...
+    function log_message {
+    echo "$(date -u) $1"
+  }
```
excieve, line 116: Indentation seems off (but then again, maybe don't put these inside the `if` body).

excieve (commented):
Thanks @abhishiekgowtham! Please take a look at a couple questions and suggestions in the comments.

Regarding templating the variables into the script as @gernest suggests, I'm not sure which approach is better.

Subjectively I lean towards templating in the pod spec and passing as env vars. The reasoning is that it makes clear what were the values of these parameters at the moment the pod was executed — don't need to inspect the script itself.

But happy to hear reasoning for templating inside the script!

tyk-pro/scripts/bootstrap_k8s.sh
abhishiekgowtham: yeah okay!

tyk-pro/scripts/bootstrap_k8s.sh
abhishiekgowtham: yes, it works without adding `http://`. I'll add it to the var anyway

tyk-pro/scripts/bootstrap_k8s.sh
abhishiekgowtham: seems to be working as expected.

tyk-pro/templates/_bootstrap_script.tpl
```diff
@@ -3,7 +3,7 @@
```
excieve, line 3: Is this file still needed?

excieve (approved):
Very nice @abhishiekgowtham!

`tyk-pro/templates/_bootstrap_script.tpl` could be removed for now I think.

abhishiekgowtham:
Thanks for the review @excieve!

PR #107, opened 2021-04-01
update the sharding information

letzya:
Remove reference to enable sharing since it's not needed any longer

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,30 +90,24 @@ The Tyk owned MDCB registry is private and requires adding users to our organisa
 
 
 
-## Caveat: Tyk license and the number of gateway nodes
+## Caveats: 
+
+### Tyk license and the number of gateway nodes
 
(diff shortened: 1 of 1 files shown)


excieve (commented):
Thanks @letzya!

Would be nice to provide some context in the PR though. I'm guessing this is due to segmentation being disabled by default now. Do we need a note on how to enable and make use of it if needed?

letzya:
I'm confused about it actually and haven't tested it so yes, it's only based on the fact we don't need the ingress tag since we do the ingress controller differently in the operator repo. 
The reason I'm confused is because if enabled and with DaemonSet in values.yaml, all the gateways have the same config and as such we will not have a control gw and afair I was always told we need one gateway to load all the apis to Redis (again have never tested to validate this assumption)

vesko-tyk:
Sharding information was updated in this PR: https://github.com/TykTechnologies/tyk-helm-chart/pull/117, hence closing this one.

PR #108, opened 2021-04-03
[TD-206] Automated smoke tests

abhishiekgowtham:
Adds a GitHub Action which runs smoke tests for PRs and master branch. Also, any suggestions on how we could test the hybrid chart?

### DIFF
--- a/.github/kind.yml
+++ b/.github/kind.yml
@@ -1,0 +1,13 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+  image: "kindest/node:v1.20.2"
+- role: control-plane
+  image: "kindest/node:v1.20.2"
+- role: control-plane
+  image: "kindest/node:v1.20.2"
+- role: worker
+  image: "kindest/node:v1.20.2"
+- role: worker
+  image: "kindest/node:v1.20.2"
--- a/.github/workflows/smoke-test.yml
+++ b/.github/workflows/smoke-test.yml
@@ -111,8 +111,7 @@ jobs:
           if [ "${{ matrix.directory }}" == tyk-headless ]; then
             helm install ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }}
           elif [ "${{ matrix.directory }}" == tyk-hybrid ]; then
-            # Need to set proper orgId, apiKey and connString values
-            helm install --set gateway.rpc.rpcKey=5cc03283d07e7f00019404b3 --set gateway.rpc.apiKey=ChAnGeMe --set gateway.rpc.connString=localhost ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }}
+            helm install --set gateway.rpc.rpcKey=${{ secrets.HYBRID_RPC_KEY }} --set gateway.rpc.apiKey=${{ secrets.HYBRID_API_KEY }} --set gateway.rpc.connString=${{ secrets.HYBRID_MDCB_HOST }} ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }}
           elif [ "${{ matrix.directory }}" == tyk-pro ]; then
             helm install --set mdcb.license="${{ secrets.MDCB_LICENSE }}" --set dash.license="${{ secrets.DASH_LICENSE }}" ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }} --wait
           fi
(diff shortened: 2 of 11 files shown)


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,127 @@
...
+          fi
+
+      - name: Upgrade "${{ matrix.directory }}" chart
+        run: |
+          if [ "${{ matrix.directory }}" == tyk-headless ]; then
+            helm upgrade -f ${{ env.TYK_CE }} ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }}
+          elif [ "${{ matrix.directory }}" == tyk-hybrid ]; then
+            helm upgrade -f ${{ env.TYK_HYBRID }} --set gateway.rpc.rpcKey=5cc03283d07e7f00019404b3 --set gateway.rpc.apiKey=ChAnGeMe --set gateway.rpc.connString=localhost ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }}
+          elif [ "${{ matrix.directory }}" == tyk-pro ]; then
+            helm upgrade -f ${{ env.TYK_PRO }} --set mdcb.license="${{ secrets.LICENSE_KEY }}" --set dash.license="${{ secrets.LICENSE_KEY }}" ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }} --wait
```
abhishiekgowtham, line 122: Need someone with higher access to add this secret for the runner.

.github/kind.yml
```diff
@@ -0,0 +1,13 @@
...
+- role: control-plane
+  image: "kindest/node:v1.20.2"
+- role: control-plane
+  image: "kindest/node:v1.20.2"
+- role: control-plane
+  image: "kindest/node:v1.20.2"
+- role: worker
+  image: "kindest/node:v1.20.2"
+- role: worker
+  image: "kindest/node:v1.20.2"
```
excieve, line 13: Is there a reason to have more than 1 CP node for tests?
Also, can it be a matrix of versions, so that we could cover different supported k8s versions (e.g. latest 1.16-1.20)?

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,127 @@
...
+        run: |
+          kubectl create namespace ${{ env.NAMESPACE }}
+          kubectl apply -f deploy/dependencies/mongo.yaml -n ${{ env.NAMESPACE }}
+          kubectl apply -f deploy/dependencies/redis.yaml -n ${{ env.NAMESPACE }}
+
+          if [ "${{ matrix.directory }}" == tyk-headless ]; then
+            helm install ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }}
+          elif [ "${{ matrix.directory }}" == tyk-hybrid ]; then
+            # Need to set proper orgId, apiKey and connString values
+            helm install --set gateway.rpc.rpcKey=5cc03283d07e7f00019404b3 --set gateway.rpc.apiKey=ChAnGeMe --set gateway.rpc.connString=localhost ${{ matrix.directory }} ./${{ matrix.directory }} -n ${{ env.NAMESPACE }} --timeout ${{ env.TIMEOUT }}
```
excieve, line 110: `rpcKey` and `apiKey` should be secrets. It will need a permanent Tyk Cloud deployment it could connect too though — please raise an internal ticket (`TD`) to set one up.

excieve (commented):
Thanks @abhishiekgowtham!

Hybrid can be done, but needs a permanent Tyk Cloud deployment for this. Alternatively, a Pro chart with MDCB enabled can be deployed and a hybrid test configured for that chart's MDCB endpoint if possible. That's slightly different than Cloud hybrid, but essentially the same thing would be tested.

Also, shouldn't we enable everything possible (portal, MDCB, etc.) to test all the things?

.github/kind.yml
abhishiekgowtham: I wanted to replicate the testing as much close to production as possible so that's why i chose to run in HA for CP.

gernest: Wont one `control-plane` and  two `worker` node satisfy minimum production  setting?

abhishiekgowtham: i don't think that's HA or really production level. But running with 5 node setup there seems to be a [bug](https://github.com/kubernetes-sigs/kind/issues/1288) in KinD so will use a 3 node setup.

abhishiekgowtham:
@excieve I have added support to run on multiple kubernetes versions. And various components like mdcb, tib, etc... are enabled during the upgrade not on install. 

I haven't enabled ingresses though.

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,132 @@
...
+      
+      - name: Install helm
+        uses: Azure/setup-helm@v1.1
+        with:
+          version: v3.5.3
+      
+      - name: Deploy "${{ matrix.directory }}" chart
+        run: |
+          kubectl create namespace ${{ env.NAMESPACE }}
+          kubectl apply -f deploy/dependencies/mongo.yaml -n ${{ env.NAMESPACE }}
```
gwotg, line 108: I'm running https://charts.bitnami.com/bitnami in a single node replicaset internally (not like in my mind). It both allows expansion to a real production mongo deployment and compatibility with our application (transaction). Should offshore the job of maintaining a deployment.

.github/workflows/smoke-test.yml
gwotg: Unblock yourself - split between what you can and cannot do without a permanent deployment, release the former and hang the later on the  ticket.

gwotg (changes requested):
I think the split up - with / without a permanent deployment - is needed before you carry on with merging - I'd even say they are distinct and failing the later shouldn't block, as the availability of that deployment isn't quite off hours pager material.

.github/workflows/smoke-test.yml
gwotg: Relevant bits from my values

`
clusterDomain: cluster.local
architecture: replicaset
useStatefulSet: true

auth:
  enabled: true
  rootPassword: "<thou shalt not>"
  replicaSetKey: ""
replicaSetName: rs0
replicaSetHostnames: true
replicaCount: 1

persistence:
  enabled: false

service:
  nameOverride: ""
  type: ClusterIP
  port: 27017
  portName: mongodb
  nodePort: ""
  externalIPs: []
  loadBalancerSourceRanges: []
  annotations: {}

externalAccess:
  enabled: true
  service:
    type: LoadBalancer
    port: 27017
    loadBalancerIPs: 
      - {{ metallb_ip_ini }}
    domain: ara.dev

arbiter:
  enabled: true
`

gwotg: Yeah, I dunno how to format yaml in a comment.

abhishiekgowtham: i don't think we need the bitnami charts as it uses StatefulSets + a bunch of other stuff. A simple deployments should suffice. However, we have redis/mongodb charts in the cloudsmith that could be used.

.github/workflows/smoke-test.yml
abhishiekgowtham: The entire hybrid chart can't be tested without permanent deployment. I mean can set dummy values like i did, but that's not the point of having smoke-tests.

I believe it's better to knock it off in a single ticket rather than coming back to it in sometime with a different ticket.

excieve (approved):
Looks good! Feel free to merge once hybrid link is available.

abhishiekgowtham:
@excieve Thanks for the approval. Merging this PR now.

joshblakeley:
@abhishiekgowtham Can we please document how we test the chart in the README?

abhishiekgowtham:
@joshblakeley Automated testing is meant for contributors so its kind of unnecessary to have it in the README. It should be part of the CONTRIBUTING.md. 

@TykTechnologies/devops what do you think?

PR #109, opened 2021-04-03
[TD-219] Tyk Pro bootstrap

abhishiekgowtham:
A couple of things i missed in #105. 

- Grant post-install hook job access to restart dashboard deployment
- Allow hooks to retry only once because if the job is retried number of times the script will bootstrap from scratch each time causing `ORG DATA: {"Status":"Error","Message":"Failed to save new Org object to DB","Meta":null}
ORG ID: null` error
- Use dashboard deployment name as an environment variable in the bootstrap script


excieve (approved):
Looks good, thanks!

PR #110, opened 2021-04-06
[TD-244] Add lint and tests

abhishiekgowtham:
This PR should allow to run additional tests (use --upgrade in ct install to test upgrading charts) and lint every pull request.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -5,6 +5,10 @@ on: pull_request
 jobs:
   lint-test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        k8s-version:
+          - v1.20.2
     steps:
       - name: Git Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


excieve (approved):
Looks good, thanks @abhishiekgowtham!

Does it depend on #108 to be merged first (Kind manifest)?

abhishiekgowtham:
Yes @excieve! Need to merge #108 first.

gwotg (approved):
Looks great @abhishiekgowtham

asoorm:
> Do we want to add any information to the secret name about expiration? Could raise awareness by just reading.

This is how we check expiry in the operator. That way the test fails if the key is expired & we are alerted via GH actions.

https://github.com/TykTechnologies/tyk-operator/blob/master/.github/workflows/pro.yaml#L38-L40

abhishiekgowtham:
Awesome, Thanks @asoorm! Will update the PR to include that step.

PR #111, opened 2021-04-07
Helm updates

zalbiraw:
- Updated all default hostname values to mirror values used in the ingress section for the respective services. 
- Changed Gateway TLS to be disabled by default.
- Added the ability to disable Dashboard bootstrapping in the tyk-pro helm.
- Turned off the `use_sharded_analytics` option in the tyk-pro helm.
- Added the ability to disable different Tyk components in the tyk-pro helm.
- Added control API gateway functionality to the tyk-pro helm.
- Fixed bootstrap vs dashboard race issues. 
- Dynamically set dashboard URL protocol based on tls option under dash service in the tyk-pro helm.

Tickets resolved:
- https://tyktech.atlassian.net/browse/TD-251

### DIFF
--- a/tyk-headless/values.yaml
+++ b/tyk-headless/values.yaml
@@ -3,34 +3,34 @@
 # Declare variables to be passed into your templates.
 
 nameOverride: ""
-fullnameOverride: "tyk-headless"
+fullnameOverride: tyk-headless
 
 secrets:
-  APISecret: "CHANGEME"
-  OrgID: "1"
+  APISecret: CHANGEME
+  OrgID: 1
 
 redis:
     shardCount: 128
     addrs:
-      - "redis.tyk.svc.cluster.local:6379"
+      - redis.tyk.svc.cluster.local:6379
     useSSL: false
     #If you're using Bitnami Redis chart please input the correct host to your installation in the field below
     #addrs:
-      #- "tyk-redis-master.tyk.svc.cluster.local:6379"
+      #- tyk-redis-master.tyk.svc.cluster.local:6379
     #If you're using Bitnami Redis chart please input your password in the field below
       #pass: ""
 
 mongo:
   enabled: false
-#    mongoURL: "mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics"
+#    mongoURL: mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics
 #    #If you're using Bitnami MongoDB chart please input your password below
-#    #mongoURL: "mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk-dashboard?authSource=admin"
+#    #mongoURL: mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk-dashboard?authSource=admin
 #    useSSL: false
 
 gateway:
   kind: DaemonSet
   replicaCount: 1
-  hostName: "tyk-gw.local"
+  hostName: tyk-gw.local
   tls: false
   containerPort: 8080
   tags: ""
@@ -53,7 +53,7 @@ gateway:
     enabled: false
     annotations: {}
     # kubernetes.io/ingress.class: nginx
-    # kubernetes.io/tls-acme: "true"
+    # kubernetes.io/tls-acme: true
     path: /
     hosts:
       - tyk-gw.local
--- a/tyk-headless/values.yaml
+++ b/tyk-headless/values.yaml
@@ -7,7 +7,7 @@ fullnameOverride: tyk-headless
 
 secrets:
   APISecret: CHANGEME
-  OrgID: 1
+  OrgID: "1"
 
 redis:
     shardCount: 128
@@ -53,7 +53,7 @@ gateway:
     enabled: false
     annotations: {}
     # kubernetes.io/ingress.class: nginx
-    # kubernetes.io/tls-acme: true
+    # kubernetes.io/tls-acme: "true"
     path: /
     hosts:
       - tyk-gw.local
--- a/tyk-headless/values.yaml
+++ b/tyk-headless/values.yaml
@@ -31,7 +31,7 @@ gateway:
   kind: DaemonSet
   replicaCount: 1
   hostName: "tyk-gw.local"
-  tls: true
+  tls: false
   containerPort: 8080
   tags: ""
   image:
--- a/tyk-headless/values.yaml
+++ b/tyk-headless/values.yaml
@@ -30,7 +30,7 @@ mongo:
 gateway:
   kind: DaemonSet
   replicaCount: 1
-  hostName: "gateway.tykbeta.com"
+  hostName: "tyk-gw.local"
   tls: true
   containerPort: 8080
   tags: ""
--- a/tyk-pro/Chart.yaml
+++ b/tyk-pro/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: Deploys Tyk Dashboard, assumes that MongoDB and Redis have already been installed
 name: tyk-pro
-version: 0.8.0
+version: 0.8.1
--- a/tyk-pro/configs/tyk_analytics.conf
+++ b/tyk-pro/configs/tyk_analytics.conf
@@ -27,7 +27,7 @@
     "portal_root_path": "/"
   },
   "home_dir": "/opt/tyk-dashboard",
-  "use_sharded_analytics": true,
+  "use_sharded_analytics": false,
   "enable_aggregate_lookups": true,
   "enable_analytics_cache": true,
   "allow_explicit_policy_id": true,
--- a/tyk-pro/scripts/bootstrap.sh
+++ b/tyk-pro/scripts/bootstrap.sh
@@ -160,8 +160,10 @@ log_ok(){
   log_message "  Ok"
 }
 
-
-main $DASHBOARD_HOSTNAME
+if [ "$BOOTSTRAP_DASHBOARD" = "true" ]
+then
+  main $DASHBOARD_HOSTNAME
+fi
 
 kubectl create secret -n ${TYK_POD_NAMESPACE} generic tyk-operator-conf \
   --from-literal "TYK_AUTH=${USER_AUTH_CODE}" \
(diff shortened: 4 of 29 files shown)


tyk-pro/values.yaml
```diff
@@ -158,17 +159,19 @@ dash:
   org:
     name: Default Org
     # Set this value to the domain of your developer portal
-    cname: "localhost"
+    cname: tyk-portal.local
 
 portal:
+  enabled: true
```
asoorm, line 165: I would suggest that enabled should be false by default. Then provide clear instruction on how to enable it.

Optimise for simplicity, performance and security. Don't make users turn things off if they don't need it, make them explicitly enable things if they do need it. But provide sensible defaults for when they do turn it on.

joshblakeley:
Thanks @zalbiraw can we please get a proper description of the changes in the PR and what the motivation is as well as links to any issues this addresses in GH/ Jira.

joshblakeley:
Also please bump the chart version

tyk-pro/values.yaml
joshblakeley: I think we actually dont need this at all? The only place it is used is in invoking the portal ingress chart. I would rather the portal exists and then whether to expose it outside the cluster is the users choice at that point.

tyk-headless/values.yaml
```diff
@@ -53,7 +53,7 @@ gateway:
     enabled: false
     annotations: {}
     # kubernetes.io/ingress.class: nginx
-    # kubernetes.io/tls-acme: "true"
+    # kubernetes.io/tls-acme: true
```
excieve, line 56: You do need to quote bools for the annotations — otherwise the value will be misinterpreted (same in the other annotations). Again, because dynamic typing and only strings are supported in the annotations, see e.g. https://github.com/kubernetes/ingress-nginx/issues/1990

Not sure if `tls-acme` is still relevant at all (`kube-lego` is long deprecated) though.

tyk-pro/configs/tyk_analytics.conf
```diff
@@ -27,7 +27,7 @@
     "portal_root_path": "/"
   },
   "home_dir": "/opt/tyk-dashboard",
-  "use_sharded_analytics": true,
+  "use_sharded_analytics": false,
```
excieve, line 30: Why? Will it affect existing deployments?

tyk-pro/scripts/bootstrap.sh
```diff
@@ -13,6 +13,9 @@ main(){
   then
     syntax_message
   else
+    # Sleep for 10 seconds to allow for dashboard to come up.
+    sleep 10
```
excieve, line 17: Better to check for the deployment status to be deterministic about this, also Helm's `--wait` should be doing this already.

excieve (changes requested):
Thanks @zalbiraw! Please check the comments, also please provide a description of changes and reasoning for them.

tyk-pro/values.yaml
zalbiraw: I agree with you @joshblakeley only reason I added this is consistency. Is that something we want to keep for that or no? Because otherwise it is not providing any value. @asoorm that sounds good to me as well if we do decide to keep it I will set it to false by default

tyk-headless/values.yaml
```diff
@@ -3,35 +3,35 @@
...
 
 nameOverride: ""
-fullnameOverride: "tyk-headless"
+fullnameOverride: tyk-headless
 
 secrets:
-  APISecret: "CHANGEME"
-  OrgID: "1"
+  APISecret: CHANGEME
+  OrgID: 1
```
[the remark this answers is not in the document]
zalbiraw: I can add the quotes back to the int values. The reason I removed them is for consistency of quotes usages.

tyk-pro/values.yaml
joshblakeley: I also disagree it should be false if you did keep it.

tyk-headless/values.yaml
zalbiraw: Ahhh good catch I will throw these back in.

tyk-pro/configs/tyk_analytics.conf
zalbiraw: This currently is breaking the analytics in the log browser. With the current pumps in the pump.conf. This only applies for multi-tenants and requires the selective pump to work. Issue is not a lot of users use multi tenancy and that is required in their licenses so we thought it is best to turn it off.

tyk-pro/scripts/bootstrap.sh
zalbiraw: I am going to replace this with retires on the /org endpoint. --wait is currently not resolving the issue. If you want to recreate the issue disable gateway installation on helm. You will run into the problem then. Retries are much more elegant however.

tyk-pro/scripts/init.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+apk --no-cache add curl jq
+curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+
+/opt/scripts/wait-for-it.sh -t 120 $TYK_DASHBOARD_SVC.$TYK_POD_NAMESPACE.svc.cluster.local:$TYK_DB_LISTENPORT && \
```
abhishiekgowtham, line 9: Do we need to ship wait-for-it with our helm charts. I think it's kinda overkill. 

This simple bash line should be able to check for the dashboard status: `while [[ $(kubectl get po -l app=dashboard-tyk-pro -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do sleep 10; done`

tyk-pro/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: "1.0"
 description: Deploys Tyk Dashboard, assumes that MongoDB and Redis have already been installed
 name: tyk-pro
-version: 0.8.0
+version: 0.8.1
```
abhishiekgowtham, line 5: I forgot to bump the version in #105 but with these many changes, shouldn't it be a minor version?

tyk-pro/scripts/init.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -e
+
+apk --no-cache add curl jq
```
abhishiekgowtham, line 5: It's good to add repo for each package as both packages are not available on main repository and v3.13 branch.

tyk-pro/scripts/init.sh
zalbiraw: We could, I normally use this script around all my projects. I don't have a problem keeping it or replacing it.

tyk-pro/Chart.yaml
zalbiraw: @excieve recommended thiS version update. But I don't mind a minor version update either.

tyk-pro/scripts/init.sh
zalbiraw: I just moved what we had in the bootstrap. If you'd like to make the change, by all means.

tyk-pro/values.yaml
```diff
@@ -13,25 +13,28 @@ enableSharding: false
...
     useSSL: false
     #If you're using Bitnami Redis chart please input your password in the field below
     #pass: ""
 
 mongo:
-    # mongoURL: "mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics"
+    # mongoURL: mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics
     #If you're using Bitnami MongoDB chart please input your password below
-    #mongoURL: "mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk-dashboard?authSource=admin"
+    #mongoURL: mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk-dashboard?authSource=admin
```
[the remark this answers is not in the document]
zalbiraw: I am not sure. But from my understanding this will treated as a string regardless no?

tyk-pro/scripts/init.sh
abhishiekgowtham: can you replace it please? Needs to be tested properly.

tyk-pro/scripts/init.sh
abhishiekgowtham: i think this can be in a different PR.

abhishiekgowtham:
@zalbiraw Except the wait-for-it part everything else LGTM!

abhishiekgowtham (approved):
@zalbiraw Except the wait-for-it part everything else LGTM!

tyk-pro/scripts/init.sh
zalbiraw: Awesome thanks. I'll resolve this then.

vesko-tyk:
In the `tyk-pro` chart, If the Dashboard bootstrapping is disabled then `NOTES.txt` file should be updated to cater only for the use cases this option is relevant. At the moment is a bit misleading as if we set dashboard bootstrap to `false` when installing the `tyk-pro` chart we're still showing the notes which says: `The Dashboard is bootstrapped automatically...`
Please update this accordingly. 
Thanks!

tyk-pro/values.yaml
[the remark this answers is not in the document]
excieve: As this is injected into an env var, lack of string quoting (e.g. in username/password) may break stuff. But I see it's already passed through a `quote` filter in the templates, so this should be fine.

tyk-pro/scripts/init.sh
excieve: @zalbiraw @abhishiekgowtham Looks like this wasn't done, but the PR was merged nonetheless. I agree with Gowtham that it's better to have a simpler (easier to reason about) code if possible, so maybe do in a separate PR?

abhishiekgowtham: @excieve i'll work on this and the alpine packages issue in a different PR.

PR #112, opened 2021-04-07
Update deployment-gw-repset.yaml

clemensk1:
fixed yaml parsing error when control api -> enabled

### DIFF
--- a/tyk-headless/templates/deployment-gw-repset.yaml
+++ b/tyk-headless/templates/deployment-gw-repset.yaml
@@ -109,7 +109,7 @@ spec:
         workingDir: /opt/tyk-gateway
         ports:
         - containerPort: {{ .Values.gateway.containerPort }}
-        {{- if .Values.gateway.control.enabled -}}
+        {{- if .Values.gateway.control.enabled }}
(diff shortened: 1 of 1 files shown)


gwotg (approved):
Thanks for chiming in @clemensk1! Good catch

tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -109,7 +109,7 @@ spec:
         workingDir: /opt/tyk-gateway
         ports:
         - containerPort: {{ .Values.gateway.containerPort }}
-        {{- if .Values.gateway.control.enabled -}}
+        {{- if .Values.gateway.control.enabled }}
```
abhishiekgowtham, line 112: @gwotg By enabling the control in the gateway, the containerPort number does not change to 9696 and still runs on 8080. It's suppose to be structured this way:

`      {{- if .Values.gateway.control.enabled }}
        - containerPort: {{ .Values.gateway.control.containerPort }}
        {{- else }}
        - containerPort: {{ .Values.gateway.containerPort }}
        {{- end }}`

PR #115, opened 2021-04-09
[TD-206] Automated smoke tests

abhishiekgowtham:
Fix some issues that are not addressed in #108.

### DIFF
--- a/ct.yaml
+++ b/ct.yaml
@@ -3,3 +3,4 @@ chart-dirs:
   - ./
 target-branch: master
 helm-extra-args: --timeout 8m
+check-version-increment: false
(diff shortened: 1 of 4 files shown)


excieve (approved):
:ok_hand:

ct.yaml
```diff
@@ -2,4 +2,5 @@ remote: origin
 chart-dirs:
   - ./
 target-branch: master
-helm-extra-args: --timeout 8m
\ No newline at end of file
+helm-extra-args: --timeout 8m
+check-version-increment: false
```
abhishiekgowtham, line 6: @excieve this should address the issue we discussed regarding bumping chart versions on every PR.

PR #119, opened 2021-04-17
Various small fixes

sedkis:
1.  Hard coding the Dashboard to http as there is no option to enable TLS. Currently, we set the Dashboard protocol based off the Gateway TLS option which makes no sense. We can turn on the Gateway TLS and this leaves the Dashboard TLS off.

2.  Adding comments to explain the API secrets


abhishiekgowtham (commented):
Thanks @sedkis! Everything looks good. I've noticed that CI pipeline is failing, can you fix that please?

PR #120, opened 2021-04-19
 [TD-40] Allow to configure redis database index through values

vesko-tyk:
Allow to configure redis database index for all charts through values.yaml file.

### DIFF
--- a/tyk-headless/values.yaml
+++ b/tyk-headless/values.yaml
@@ -10,10 +10,10 @@ secrets:
   OrgID: "1"
 
 redis:
-    shardCount: 128
-    addrs:
-      - redis.tyk.svc.cluster.local:6379
-    useSSL: false
+  shardCount: 128
+  addrs:
+    - redis.tyk.svc.cluster.local:6379
+  useSSL: false
 #    #If you're using Bitnami Redis chart please input the correct host to your installation in the field below
 #    addrs:
 #      - tyk-redis-master.tyk.svc.cluster.local:6379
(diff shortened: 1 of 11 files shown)


tyk-headless/values.yaml
```diff
@@ -14,11 +14,16 @@ redis:
     addrs:
       - redis.tyk.svc.cluster.local:6379
     useSSL: false
-    #If you're using Bitnami Redis chart please input the correct host to your installation in the field below
-    #addrs:
-      #- tyk-redis-master.tyk.svc.cluster.local:6379
-    #If you're using Bitnami Redis chart please input your password in the field below
-      #pass: ""
+#    #If you're using Bitnami Redis chart please input the correct host to your installation in the field below
```
abhishiekgowtham, line 17: I think it'd better to address other managed redis services here too rather than redis bitnami chart.

abhishiekgowtham (commented):
Thanks @vesko-tyk! I have left a single comment on the changes. 

Also, looks the ci pipeline is failing. Can you fix that please?

tyk-headless/values.yaml
vesko-tyk: @abhishiekgowtham Can you elaborate a bit more on that? The only managed redis service we recommend to our user based on the docs at the moment is from Bitnami, not sure what other managed redis services you're referring to here?

vesko-tyk: Cool, got it now and what you're saying make sense. However this change is out of scope of this PR and I believe it will require a bit more work than just changing a comment inside the values file eg. proper testing of those options you're suggesting, updating the docs with some examples etc.

abhishiekgowtham: Sure, if you feel like it's out of the scope for this one, we can tackle that in a different PR.

abhishiekgowtham:
LGTM!

abhishiekgowtham (approved):
LGTM!

PR #121, opened 2021-04-20
[TD-42] Store sensitive info as secrets

vesko-tyk:


tyk-headless/templates/secrets.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: secrets
```
excieve, line 4: This should be more in line with the rest of the resources.
```suggestion
  name: secrets-{{ include "tyk-headless.fullname" . }}
```

Same in the other charts.

tyk-headless/templates/secrets.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: Secret
+metadata:
```
excieve, line 3: Please add the standard set of labels too.

Same in the other charts.

tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -86,13 +86,22 @@ spec:
           - name: TYK_GW_STORAGE_ENABLECLUSTER
             value: "{{ default "false" .Values.redis.enableCluster }}"
           - name: TYK_GW_STORAGE_PASSWORD
-            value: "{{ .Values.redis.pass }}"
+            valueFrom:
+              secretKeyRef:
+                name: secrets
```
excieve, line 91: I suggest making it possible to optionally specify a secret name as a value in case user doesn't want to store plaintext secrets in the Helm values and would rather provide the k8s `Secret` externally.

Same in the other charts.

tyk-headless/templates/secrets.yaml
```diff
@@ -0,0 +1,10 @@
...
+  name: secrets
+  namespace: {{ .Release.Namespace }}
```
excieve, line 5: Explicitly specifying the `namespace` isn't needed — Helm will deploy all the namespaced resources into the same NS.

excieve (changes requested):
Thanks @vesko-tyk! Looks great in general, but please add more metadata to the secrets and I'd suggest to make it possible to specify the secret name as a value optionally to let the user provide secrets externally.

Also, please check the CI fails.

tyk-headless/templates/secrets.yaml
vesko-tyk: Labels are now added.

tyk-headless/templates/secrets.yaml
vesko-tyk: Namespace is now removed.

tyk-headless/templates/deployment-gw-repset.yaml
vesko-tyk: @excieve Could you give some examples of what is needed here? Thanks!

tyk-headless/templates/secrets.yaml
vesko-tyk: Have changed the name now.

tyk-headless/templates/deployment-gw-repset.yaml
vesko-tyk: @excieve I have added option for user to provide the k8s `Secret` externally. Please review it when you have a moment. Thanks!

tyk-headless/templates/deployment-gw-repset.yaml
```diff
@@ -88,13 +88,28 @@ spec:
           - name: TYK_GW_STORAGE_DATABASE
             value: "{{ default "0" .Values.redis.storage.database }}"
           - name: TYK_GW_STORAGE_PASSWORD
-            value: "{{ .Values.redis.pass }}"
+            {{- if .Values.SecretName.enabled -}}
+            valueFrom:
+              secretKeyRef:
+                name: secrets-{{ include "tyk-headless.fullname" . }}
+                key: redisPass
+            {{- end }}
```
excieve, line 96: This leads to a broken `env` definition since  when `SecretName` is not enabled then this env var has no value. Also I think logic is wrong here — when a user wants to provide the secret name, it's not actually possible without modifying the charts itself.

So maybe something along the lines of:
```.yaml
- name: TYK_GW_STORAGE_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ if .Values.secrets.useSecretName }} {{ .Values.secrets.useSecretName }} {{ else }} secrets-{{ include "tyk-headless.fullname" . }} {{ end}}
      key: redisPass
```

Or something more pretty with a helper template call.

Same everywhere else.

tyk-headless/templates/secrets.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.SecretName.enabled -}}
```
excieve, line 1: To align with the other comments:
```suggestion
{{- if not .Values.secrets.useSecretName -}}
```

tyk-headless/values.yaml
```diff
@@ -9,6 +9,10 @@ secrets:
   APISecret: CHANGEME
   OrgID: "1"
 
+# If you don't want to store plaintext secrets in the Helm value file and would rather provide the k8s Secret externally please disable this value
+SecretName:
```
excieve, line 13: I'd suggest to perhaps place it under the `secrets` var and use lowercase first word as convention goes for the Helm values, e.g.:

```.yaml
secrets:
  useSecretName: ""
```

excieve (changes requested):
Thanks @vesko-tyk! As discussed let's make sure this allows user to actually provide a secret name (as an option), but default to generate if empty.

vesko-tyk:
@excieve Thanks for the review and very helpful suggestions. I believe all issues related to the implementation of the secrets have been fixed now.

However there is small caveat with the CI tests - the Tyk Pro helm chart is failing and I can't figure out why. If I install the chart on my local everything is working fine without issues. Bootstrap is successful, I can access my dashboard, create API's, keys etc.

abhishiekgowtham (changes requested):
I don't know why #122 was merged into this PR since it hasn't been merged into master yet.

tyk-pro/values.yaml
```diff
@@ -37,7 +39,7 @@ redis:
     database: 0
 
 mongo:
-  # mongoURL: mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics
+  mongoURL: mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics
```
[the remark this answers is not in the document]
excieve: Didn't help apparently.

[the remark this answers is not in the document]
excieve: I think I know why — the new secret doesn't use the mongo url helper anymore (which was setting a correct value). Will try using it again.

excieve:
@abhishiekgowtham @gwotg Tests are now passing :fireworks: after https://github.com/TykTechnologies/tyk-helm-chart/pull/121/commits/7250697111135b8edacb552c89367a7cd7ca0724

tyk-pro/templates/deployment-dash.yaml
```diff
@@ -67,9 +67,9 @@ spec:
           - name: TYK_DB_HOSTCONFIG_GATEWAYHOSTNAME
             value: "{{ .Values.gateway.hostName }}"
           {{- if .Values.gateway.control.enabled }}
-          - name: TYK_DB_TYKAPI_HOST
+          - name: TYK_GW_CONTROLAPIHOSTNAME
             value: "{{ include "tyk-pro.gwproto" . }}://gateway-control-svc-{{ include "tyk-pro.fullname" . }}.{{ .Release.Namespace }}"
-          - name: TYK_DB_TYKAPI_PORT
+          - name: TYK_GW_CONTROLAPIPORT
             value: "{{ .Values.gateway.control.containerPort }}"
```
excieve, line 73: Hmm, this is wrong. Looks like a merge didn't go well somewhere.

tyk-pro/templates/deployment-mdcb.yaml
```diff
@@ -115,6 +120,10 @@ spec:
           periodSeconds: 10
           timeoutSeconds: 3
           failureThreshold: 3
+      {{- with .Values.mdcb.image.pullSecrets }}
+      imagePullSecrets:
+      {{- toYaml . | nindent 12 }}
```
excieve, line 125: While this change is a really good catch, not sure there are still 12 units to indent (gotta love yaml!)

excieve (changes requested):
A couple more irregularities.

excieve (approved):
Thanks!

tyk-pro/templates/deployment-mdcb.yaml
gwotg: Thanks @excieve. I'm all tested here and think we're good to go. Resolving both

.github/workflows/smoke-test.yml
```diff
@@ -7,7 +7,7 @@ on:
       - master
 
 env:
-  TIMEOUT: 3m
+  TIMEOUT: 10m
```
abhishiekgowtham, line 10: this seems quite long. Can we revert it back to 3m please? Usually, we don't need to edit this value, if the chart isn't available after this time which means there's an error or chart is unstable.

abhishiekgowtham (approved):
/lgtm

excieve:
Thanks all!

PR #122, opened 2021-04-22
[TD-283] liveness and readiness probe to query control api

abhishiekgowtham:
Allow gateway liveness and readiness probe to query control-api when it's enabled. Also, the `/hello` endpoint might not be available on control api for specific versions so it checks for that and update the port accordingly.

### DIFF
--- a/.github/tyk_pro_values.yaml
+++ b/.github/tyk_pro_values.yaml
@@ -21,21 +21,26 @@ secrets:
   AdminSecret: "12345"
 
 redis:
-    shardCount: 128
-    #If you're using Bitnami Redis chart please input the correct host to your instalaltion in the field below
-    #addrs:
-    #   - tyk-redis-master.tyk.svc.cluster.local:6379
-    # addrs:
-    #   - redis.tyk.svc.cluster.local:6379
-    useSSL: false
-    #If you're using Bitnami Redis chart please input your password in the field below
-    #pass: ""
+  shardCount: 128
+  # addrs:
+  #   - redis.tyk.svc.cluster.local:6379
+  useSSL: false
+#    #If you're using Bitnami Redis chart please input the correct host to your installation in the field below
+#    addrs:
+#      - tyk-redis-master.tyk.svc.cluster.local:6379
+#   #If you're using Bitnami Redis chart please input your password in the field below
+#    pass: ""
+#   #If you are using Redis cluster, enable it here.
+#    enableCluster: false
+#   #By default the database index is 0. Setting the database index is not supported with redis cluster. As such, if you have enableCluster: true, then this value should be omitted or explicitly set to 0.
+  storage:
+    database: 0
 
 mongo:
-    # mongoURL: mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics
-    #If you're using Bitnami MongoDB chart please input your password below
-    #mongoURL: mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk-dashboard?authSource=admin
-    useSSL: false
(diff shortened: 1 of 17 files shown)


abhishiekgowtham:
@excieve All tests have passed and it can merged for v0.9.0 release.

tyk-pro/templates/deployment-dash.yaml
```diff
@@ -71,9 +71,9 @@ spec:
           - name: TYK_DB_TYKAPI_PORT
             value: "{{ .Values.gateway.service.port }}"
           {{- if .Values.gateway.control.enabled }}
-          - name: TYK_DB_TYKAPI_HOST
+          - name: TYK_GW_CONTROLAPIHOSTNAME
             value: "{{ include "tyk-pro.gwproto" . }}://gateway-control-svc-{{ include "tyk-pro.fullname" . }}.{{ .Release.Namespace }}"
-          - name: TYK_DB_TYKAPI_PORT
+          - name: TYK_GW_CONTROLAPIPORT
```
abhishiekgowtham, line 76: environment variables are wrong based on the [docs](https://tyk.io/docs/tyk-oss-gateway/configuration/#control_api_hostname). Smoke tests were failing during upgrade due to this error.

abhishiekgowtham, line 76: environment variables are wrong based on the [docs](https://tyk.io/docs/tyk-oss-gateway/configuration/#control_api_hostname). Smoke tests were failing during upgrade due to this error.

excieve: @abhishiekgowtham This spec is for the dashboard, not the gateway. `TYK_GW_` prefix is no-op on the dashboard, so these variables will do nothing here.

The previous variables are correct in this context. What seems to be wrong is that the condition block is incorrect — if `gateway.control.enabled` is `true` then the variables will be specified twice and the first one will likely win out.

.github/tyk_pro_values.yaml
```diff
@@ -6,44 +6,52 @@ nameOverride: ""
...
-    #mongoURL: "mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk-dashboard?authSource=admin"
-    useSSL: false
+  # mongoURL: mongodb://mongo.tyk.svc.cluster.local:27017/tyk_analytics
+  # If you're using Bitnami MongoDB chart please input your password below
+  # mongoURL: mongodb://root:pass@tyk-mongo-mongodb.tyk.svc.cluster.local:27017/tyk-dashboard?authSource=admin
+  useSSL: false
 
 mdcb:
-  enabled: true
+  enabled: false
```
excieve, line 46: Why disable it and other components for the tests?

excieve (changes requested):
@abhishiekgowtham This looks good, but the gw control API env variables on the dashboard deployment spec seem to be wrong.

.github/tyk_pro_values.yaml
abhishiekgowtham: I've enabled it now.

tyk-pro/templates/deployment-dash.yaml
abhishiekgowtham: @excieve Addressed. But having the same environment variables enabled [fails](https://github.com/TykTechnologies/tyk-helm-chart/pull/122/checks?check_run_id=2681563569#step:6:16) during an helm upgrade. The dashboard should speak to control api when `gateway.control.enabled` is enabled, if not connect to the default `hostname` and `port`.

excieve: @abhishiekgowtham Correct, so there's a block that duplicates these env vars, which was likely causing issues. I see you corrected it now with an `else` clause and the tests are passing. So this seems to be fixed.

excieve (approved):
LGTM!

PR #123, opened 2021-04-26
For future debugging

alephnull:
You must create a secret named ACTIONS_STEP_DEBUG with the value true to see the debug messages set by this command in the log.


abhishiekgowtham (approved):
LGTM!

PR #125, opened 2021-05-06
Update Docker image version to latest stable

vesko-tyk:


abhishiekgowtham (commented):
Thanks @vesko-tyk! Everything looks good except the ci fail tests. Also, do you think it's good to update readme or add comments about version upgrade?

excieve:
Might need to wait on dash 3.2.1/all 3.2.2 as 3.2.0 is a bit weird. @alephnull could you please confirm?

PR #126, opened 2021-05-12
allow more flexible expression of extraEnv

drpebcak:
This allows things like setting vars from configmaps/secrets/downward api/etc.

### DIFF
--- a/tyk-pro/templates/deployment-dash.yaml
+++ b/tyk-pro/templates/deployment-dash.yaml
@@ -119,12 +119,9 @@ spec:
                 key: DashLicense
           - name: TYK_DB_HOSTCONFIG_PORTALROOTPATH
             value:  {{ .Values.portal.path }}
-        {{- if .Values.dash.extraEnvs }}
-        {{- range $env := .Values.dash.extraEnvs }}
-          - name: {{ $env.name }}
-            value: {{ $env.value | quote }}
-        {{- end }}
-        {{- end }}
+          {{- if .Values.dash.extraEnvs }}
+          {{- toYaml .Values.dash.extraEnvs| nindent 10 }}
+          {{- end }}
         resources:
 {{ toYaml .Values.dash.resources | indent 12 }}
         command: ["/opt/tyk-dashboard/tyk-analytics", "--conf=/etc/tyk-dashboard/tyk_analytics.conf"]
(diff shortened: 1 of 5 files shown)


abhishiekgowtham (changes requested):
Thanks for improving the helm charts @drpebcak! Can you fix the failing tests please? Other than it lgtm.

drpebcak:
@abhishiekgowtham I don't think these test failures have anything to do with my changes. The `extraEnvs` change I am making has no default values in the chart.. so the rendered manifests are the same now as they were before. The output from the github jobs is not really super helpful either, but it seems like something unrelated is going wrong in the post-install script.

abhishiekgowtham:
@drpebcak The CI tests runs on every PR by explicitly checking out the branch (on the left side) so all the checks run against the changes made on that particular branch. If the output/logs on the tests isn't super helpful, I'd suggest adding some debug statements to the actions file to get some useful information.

I'm sorry, I can't accept this PR without all checks passing.

asoorm:
> @drpebcak The CI tests runs on every PR by explicitly checking out the branch (on the left side) so all the checks run against the changes made on that particular branch. If the output/logs on the tests isn't super helpful, I'd suggest adding some debug statements to the actions file to get some useful information.
> 
> I'm sorry, I can't accept this PR without all checks passing.

Screenshot 2021-06-09 at 07 39 37

@abhishiekgowtham this is our bad. If we look at the test that fails, it is caused by a license expiry check. This pulls a GH secret, and because @drpebcak is not in the Tyk GitHub organization, the GH action does not have enough permissions to pull the Tyk License to validate it's expiry. As such, the test fails.

IMO, this is a big problem, because it means that nobody in the OS community can collaborate with this project.

excieve:
@asoorm Thanks for bringing it up! We agreed to stop requiring the smoke tests for `tyk-pro` and `tyk-headless` for now, while keeping them on the `master` pushes. In the future we'll need to find a way to test these without having to require secrets. I don't think we can currently successfully test the Tyk Dashboard bootstrap without a valid license or Hybrid without a valid set of RPC creds.

asoorm:
@excieve we have similar issue with Tyk Operator. The way we "worked around" the problem was for 2 actions to fire on every PR. One for Open Source which doesn't need secrets (tyk license). And the other which needs the license. When reviewing an external PR from a fork, the Open Source action should pass, and the one which needed the license fails at the license check step, we then can manually check/test & if it looks OK, we can merge. It's not ideal, but it kind of works for the short-term. If you find a better longer term solution, please advise & we can update tyk-operator CI also.

excieve:
@asoorm I think it's a fundamental issue of running potentially untrusted code in an environment with secrets.

excieve:
@drpebcak We've pushed some CI changes to master. Could you please merge them into your branch?

drpebcak:
@excieve done!

excieve (approved):
Thanks @drpebcak!

PR #132, opened 2021-06-14
Fix dashboard HTTPS and OSS control port hard code

sedkis:
remove control port from config, its being set by env file, and unset the dashboard https based off gateway

### DIFF
--- a/tyk-headless/configs/tyk_mgmt.conf
+++ b/tyk-headless/configs/tyk_mgmt.conf
@@ -1,6 +1,5 @@
 {
     "listen_port": 8080,
-    "control_api_port": 9696,
     "secret": "$sharedsecret",
     "node_secret": "$sharedsecret",
     "template_path": "/opt/tyk-gateway/templates",
--- a/tyk-pro/templates/_helpers.tpl
+++ b/tyk-pro/templates/_helpers.tpl
@@ -41,7 +41,7 @@ http
 
 {{- define "tyk-pro.dash_proto" -}}
 {{- if .Values.gateway.tls -}}
-https
+http
 {{- else -}}
 http
 {{- end -}}


tyk-headless/configs/tyk_mgmt.conf
```diff
@@ -1,6 +1,5 @@
 {
     "listen_port": 8080,
-    "control_api_port": 9696,
```
excieve, line 3: Is this compatible with the default values (i.e. if someone upgrades to this, would their control port change)?

tyk-pro/templates/_helpers.tpl
```diff
@@ -41,7 +41,7 @@ http
 
 {{- define "tyk-pro.dash_proto" -}}
 {{- if .Values.gateway.tls -}}
-https
+http
```
excieve, line 44: This looks weird. Could you please elaborate?

tyk-headless/configs/tyk_mgmt.conf
sedkis: Yes it would.  But I think it is a necessary breaking change because it contradicts what's in the values.yaml

if tyk-headless/values.yaml `control-port.enabled` is set to `false`, it will still get set because of the hard core above

tyk-pro/templates/_helpers.tpl
sedkis: Yeah this is a flat out mistake.  in the above logic, if you set `gateway.tls=true` then it sets the dashboard url to `https` but NOT the dashboard itself, because that is controled by `dashboard.tls`

excieve: We'll merge this for now, but the entire `tyk-pro.dash_proto` helper is then pointless and should be removed I think.

excieve (approved):
Thanks! The `tyk-pro.dash_proto` should probably be removed later as it's not doing anything now.

abhishiekgowtham:
@excieve Smoke tests are failing on master for tyk-pro after this PR was merged. At least, we need to update the README to let users know that they shouldn't be cloning master branch and use latest release tag.

PR #133, opened 2021-06-18
[TD-422] Allow users to mount files as secrets

zalbiraw:
Add ability to mount files as secrets to all the Tyk stack components in all the charts

### DIFF
--- a/tyk-headless/values.yaml
+++ b/tyk-headless/values.yaml
@@ -119,6 +119,10 @@ gateway:
       effect: NoSchedule
   affinity: {}
   extraEnvs: []
+  mounts: []
+    # - name: mount
+    #   filename: mount.txt
+    #   mountPath: /dev/null
 
 # If pump is enabled the Gateway will create and collect analytics data to send
 # to a data store of your choice. These can be set up in the pump config. The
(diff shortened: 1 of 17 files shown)


abhishiekgowtham:
Hey @zalbiraw! I've made some changes to the PR, please review them. If everything looks good, we can merge.

tyk-headless/values.yaml
```diff
@@ -150,5 +154,9 @@ pump:
   tolerations: []
   affinity: {}
   extraEnvs: []
+  mounts: []
+    # - name: mount
+    #   filename: mount.txt
+    #   mountPath: /dev/null
```
abhishiekgowtham, line 160: I think it's better to have a single generic example on how to mount files which can be used for all components. I don't think we need to do this for each component. 

Same thing for all other charts too.

abhishiekgowtham (changes requested):
Thanks @zalbiraw! Please update the files based on the feedback.

abhishiekgowtham (approved):
/lgtm

@vesko-tyk Please review again before merging.

vesko-tyk (approved):
Looks good, thanks!

PR #136, opened 2021-06-24
Update README.md

marksou:
Added note about the use of kubens utility in some commands in the video.


vesko-tyk:
With new release 0.9.0 coming up, the way we install the charts would change, so this video would be no longer up to date.

PR #138, opened 2021-06-30
[TD-256] Replace wait-for-it script

abhishiekgowtham:
I think having wait-for-it bash script packaged with tyk-pro chart is kinda an overkill. We've discussed it in #111, therefore, i've replaced it with a simple `kubectl` command which waits for the dashboard pod before running the post-install hook.

Also, i didn't find any problem installing alpine packages in the script so didn't pin down to a specific version/branch.


vesko-tyk (approved):
Thanks @abhishiekgowtham lgtm\

Issue #139, opened 2021-07-01
When using values.redis.addr, stray newline prevents last entry from working


mjpieters:
The templates for `tyk-*.redis_url` add a stray `\n` newline character when `redis.addr` is set.

e.g. values containing:

```
redis:
  addrs:
  - redis-server-1:6379
  - redis-server-2:6379
```
results in

```
          - name: TYK_GW_STORAGE_ADDRS
            value: "redis-server-1:6379,redis-server-2:6379\n"
```

The stray newline breaks that entry. The newline comes from the `{{/* Adds support for ... }}` comment entries:

https://github.com/TykTechnologies/tyk-helm-chart/blob/aa9eee1d486858fa336b849b610ab2cfe3298349/tyk-pro/templates/_helpers.tpl#L58-L67

https://github.com/TykTechnologies/tyk-helm-chart/blob/aa9eee1d486858fa336b849b610ab2cfe3298349/tyk-headless/templates/_helpers.tpl#L41-L50

https://github.com/TykTechnologies/tyk-helm-chart/blob/aa9eee1d486858fa336b849b610ab2cfe3298349/tyk-hybrid/templates/_helpers.tpl#L42-L51

It can be fixed by replacing `{{/*` with `{{- /*`.

christtyk:
@zalbiraw mind having a look?

vesko-tyk:
@mjpieters Thank you for reporting this issue. This is now fixed in the latest release of our helm charts v0.9.1.

PR #147, opened 2021-07-08
Fix adding unnecessary new line to the end of entries

vesko-tyk:
Fixing issue: https://github.com/TykTechnologies/tyk-helm-chart/issues/139


excieve (approved):
Thanks @vesko-tyk! Some of these seem to be redundant (e.g. the comments on top of the defines as they're not affecting the values), but it won't harm I guess :)

PR #148, opened 2021-07-09
[BUG] Fix broken smoke tests

abhishiekgowtham:
After merging #138, smoke tests on tyk-pro chart are failing on master due to dashboard pod is taking more than allocated time to reach ready status. I've bumped the timeout, which should fix the broken tests.


abhishiekgowtham:
closing, since all workflows are successful after re-run.

PR #150, opened 2021-07-14
Update instructions and appVersion

letzya:
1. add appVersion
2. remove the `mongo: {}` in tyk-pro/values.yaml since it's confusing. If you set the mongoURL because of the few comment lines you don't see that there're `{}` after `mongo` that needs to be deleted. Adding a single field with the default value sort this out
3. Wrong field name. We had "tag" for sharding the gateways instead of "tags". I tested it by copying the gateway section in values.yaml and deployment and service and saw that with "tags it works.

### DIFF
--- a/tyk-headless/Chart.yaml
+++ b/tyk-headless/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "1.0"
 name: tyk-headless
+appVersion: "v3.2.1"
 version: 0.9.1
 icon: https://tyk.io/wp-content/uploads/2021/07/tyk_logo_bg.jpg
 home: https://tyk.io/
@@ -17,7 +17,7 @@ keywords:
   - open source gateway
   - edge gateway
 description: |-
-  This chart deploys the open source Tyk Gateway. It assumes Redis has already been installed.
-  Tyk is an open source Enterprise API Gateway, supporting REST, GraphQL, TCP and gRPC protocols.
-  Tyk Gateway is provided ‘Batteries-included’, with no feature lockout. Enabling your organization to control who accesses your APIs, when they access, and how they access it.
-  Tyk Technologies uses the same API Gateway for all it’s applications. Protecting, securing, and processing APIs for thousands of organizations and businesses around the world.
+  This chart deploys the open source Tyk Gateway.
+  Tyk Gateway is a fully open source Enterprise API Gateway, supporting REST, GraphQL, TCP and gRPC protocols.
+  Tyk Gateway is provided ‘Batteries-included’, with no feature lockout. It enables organizations and businesses around
+  the world to protecte, secure, and processe APIs and well as review and audit the consumed apis.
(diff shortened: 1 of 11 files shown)


tyk-headless/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
-appVersion: "1.0"
 name: tyk-headless
+appVersion: "v3.2.1"
```
abhishiekgowtham, line 3: IIRC, we've decided to remove `appVersion` altogether since it can't track all the components.

tyk-headless/Chart.yaml
```diff
@@ -17,7 +17,7 @@ keywords:
...
   - edge gateway
 description: |-
-  This chart deploys the open source Tyk Gateway. It assumes Redis has already been installed.
-  Tyk is an open source Enterprise API Gateway, supporting REST, GraphQL, TCP and gRPC protocols.
-  Tyk Gateway is provided ‘Batteries-included’, with no feature lockout. Enabling your organization to control who accesses your APIs, when they access, and how they access it.
-  Tyk Technologies uses the same API Gateway for all it’s applications. Protecting, securing, and processing APIs for thousands of organizations and businesses around the world.
+  This chart deploys the open source Tyk Gateway.
+  Tyk Gateway is a fully open source Enterprise API Gateway, supporting REST, GraphQL, TCP and gRPC protocols.
+  Tyk Gateway is provided ‘Batteries-included’, with no feature lockout. It enables organizations and businesses around
+  the world to protecte, secure, and processe APIs and well as review and audit the consumed apis.
```
abhishiekgowtham, line 23: maybe a typo here?

tyk-headless/README.md
```diff
@@ -1,7 +1,21 @@
-## Tyk Community Edition
-Provides a Tyk Installation for Community Edition Gateways
+## Tyk OSS Gateway
```
abhishiekgowtham, line 1: Since the naming has changed. Maybe it's better to change the directory to `tyk-ce` or `tyk-oss`

this may require changes to GitHub Actions.

abhishiekgowtham (changes requested):
Thanks @letzya! Overall looks good, but requires a few changes.

tyk-headless/Chart.yaml
letzya: I don't about it.  where was this decision published? regardless atm it's "1.0" and it's wrong. continue in slack

tyk-headless/README.md
letzya: we will but only in v1.0 since it's a breaking change for current users.

letzya:
@abhishiekgowtham please resolve the comment if you see fit :)

tyk-headless/Chart.yaml
abhishiekgowtham: I vaguely remember that it was made during our planning session. Maybe @excieve can provide more input on it?

As mentioned on slack, I'm inclined to remove appVersion in the Chart.yaml for all charts.

abhishiekgowtham (approved):
/lgtm

abhishiekgowtham (approved):
/lgtm

PR #153, opened 2021-08-12
Change tyk hybrid command to use the local config file isntead of def…

sedkis:
k8s hybrid deployment is not using the file that is being mounted at all.  the config file is mounted to opt/tyk-gateway and the gateway is using /etc/tyk-gateway

```
        command: ["/opt/tyk-gateway/tyk", "--conf=/etc/tyk-gateway/tyk.conf"]
```


abhishiekgowtham:
Closing this as #154 includes the change.

PR #161, opened 2021-09-28
chore: add parameterizable nodePort for gateway and dashboard

ermirizio:
I found this changes useful while deploying tyk-pro using KIND. I also added small piece of readme for guidance purposes on how to configure kind for reaching the nodePorts.

### DIFF
--- a/tyk-pro/README.md
+++ b/tyk-pro/README.md
@@ -45,3 +45,28 @@ Please refer to Tyk official docs for details about installation and setting [*T
 [*Tyk Developer portal*](https://tyk.io/docs/tyk-self-managed/tyk-helm-chart/#tyk-developer-portal),
 [*MDCB*](https://tyk.io/docs/tyk-self-managed/tyk-helm-chart/#installing-tyk-self-managed-control-plane) and
 [*Istio*](https://tyk.io/docs/tyk-self-managed/istio/).
+
+### Kind as local k8s cluster
+If you decide to use [*KIND*](https://kind.sigs.k8s.io/) as local k8s cluster you can rely on the following configuration in order to reach Tyk Gateway and Tyk Dashboard:
+
+    cat << EOF > kind-config.yaml
+    kind: Cluster
+    name: tyk-pro-cluster
+    apiVersion: kind.x-k8s.io/v1alpha4
+    nodes:
+    - role: control-plane
+      extraPortMappings:
+      - containerPort: 30001
+        hostPort: 3000
+        listenAddress: "0.0.0.0"
+        protocol: TCP
+      - containerPort: 30002
+        hostPort: 8080
+        listenAddress: "0.0.0.0"
+        protocol: TCP
+    EOF
+
+When creating the cluster please specify the config file as following:
+    kind create cluster --config=kind-config.yaml
+
+Then you can access *Tyk-Gateway* on http://localhost:13000 and *Tyk-Dashboard* on http://localhost:18080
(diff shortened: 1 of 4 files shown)


tyk-pro/README.md
```diff
@@ -45,3 +45,28 @@ Please refer to Tyk official docs for details about installation and setting [*T
...
+        listenAddress: "0.0.0.0"
+        protocol: TCP
+      - containerPort: 30002
+        hostPort: 8080
+        listenAddress: "0.0.0.0"
+        protocol: TCP
+    EOF
+
+When creating the cluster please specify the config file as following:
+    kind create cluster --config=kind-config.yaml
```
abhishiekgowtham, line 70:
```suggestion
   ```shell
    kind create cluster --config=kind-config.yaml
    ```
```

tyk-pro/README.md
```diff
@@ -45,3 +45,28 @@ Please refer to Tyk official docs for details about installation and setting [*T
...
+
+Then you can access *Tyk-Gateway* on http://localhost:13000 and *Tyk-Dashboard* on http://localhost:18080
```
abhishiekgowtham, line 72:
```suggestion
Then you can access *Tyk-Gateway* on `http://localhost:13000` and *Tyk-Dashboard* on `http://localhost:18080`
```

tyk-pro/values.yaml
```diff
@@ -213,6 +213,7 @@ dash:
   service:
     type: NodePort
     port: 3000
+    nodePort: 30001
```
abhishiekgowtham, line 216: I think having this set to a particular port would affect the service that are not running in a KinD cluster. Maybe move this whole section into an example directory and link that in the `README.md`

tyk-pro/README.md
```diff
@@ -45,3 +45,28 @@ Please refer to Tyk official docs for details about installation and setting [*T
...
+
+    cat << EOF > kind-config.yaml
+    kind: Cluster
+    name: tyk-pro-cluster
+    apiVersion: kind.x-k8s.io/v1alpha4
+    nodes:
+    - role: control-plane
+      extraPortMappings:
+      - containerPort: 30001
+        hostPort: 3000
```
abhishiekgowtham, line 60: It's better to pick a port that's in 60k+ range so there wouldn't any conflict.

tyk-pro/README.md
```diff
@@ -45,3 +45,28 @@ Please refer to Tyk official docs for details about installation and setting [*T
...
+        listenAddress: "0.0.0.0"
+        protocol: TCP
+      - containerPort: 30002
+        hostPort: 8080
```
abhishiekgowtham, line 64: It's better to pick a port that's in 60k+ range so there wouldn't any conflict.

tyk-pro/values.yaml
```diff
@@ -294,6 +295,7 @@ gateway:
   service:
     type: NodePort
     port: 8080
+    nodePort: 30002
```
abhishiekgowtham, line 298: Similar comment to the change in dashboard.

abhishiekgowtham (changes requested):
Overall looks good, just some nitpicks.

tyk-pro/templates/service-dash.yaml
```diff
@@ -12,6 +12,9 @@ spec:
   ports:
   - port: {{ .Values.dash.service.port }}
     targetPort: {{ .Values.dash.containerPort }}
+    {{- if .Values.dash.service.nodePort }}
```
excieve, line 15: `nodePort` is only valid if the `type` of Service is either `NodePort` or `LoadBalancer`. Perhaps worth specifying the condition accordingly to avoid cases of misconfiguration?

letzya:
@ermirizio Hi, nice and useful idea to improve DX for dev env. Would you want to update the PR base on the comments above? THanks

PR #163, opened 2021-09-30
fix a typo

kulong0105:


abhishiekgowtham:
Thanks for the contribution @kulong0105! There's a similar bug in the`tyk-hybrid` chart. Would you be willing to fix that as well?

abhishiekgowtham (commented):
Thanks for the contribution @kulong0105! There's a similar bug in the `tyk-hybrid` chart. Would you be willing to fix that as well?

abhishiekgowtham:
@excieve @vesko-tyk Need to cut a patch release.

PR #165, opened 2021-10-11
tyk-headless: add service port name to fix ingress issue

kulong0105:
the ingress-gw.yaml file use 'servicePort: http' to config service port, so need to add service port name in 'service' config


kulong0105:
ping @abhishiekgowtham

vesko-tyk:
@kulong0105 with this PR: https://github.com/TykTechnologies/tyk-helm-chart/pull/184 we no longer set `servicePort: http` in the `ingress-gw.yaml` file and instead this can be modified in the values file.

PR #172, opened 2021-10-25
Tyk headless now uses a StatefulSet and PVC for state persistence

### LINKED ISSUE #170: The default configuration for tyk-headless has no persistence and no API consistency
The headless chart creates the api gateways as a DaemonSet with emptyDir storage. This means multiple gateways pods, each with their own independent view of state. When the API is used to create apps via tyk-sync, tyk-operator, or manually with curl the result is a mess of inconsistent state depending upon which pod requests are routed to. Additionally, when a pod is destroyed and recreated all of its state disappears.

I would suggest a short term fix of switching the headless chart to default to a `Deployment` with `replicaCount=1` and add a PVC for storage rather than an emptyDir. This would make the default install function correctly, although with scale limited to one pod.

The open source edition of Tyk seems to be architecturally limited to a single instance of the gateway pod without some additional "config has changed" events or hooks to trigger config reloads on all pods.

seanhoughton:
This change fixes #170 by limiting the number of API pods to one and placing state on a persistent volume instead of ephemeral emptyDisk.

The chart also allows a pre-defined PVC to be used if shared storage is available although that would also require manually forcing a config reload on all pods whenever the configuration state changes which is not implemented in any way with this PR.

### DIFF
--- a/tyk-headless/templates/deployment-gw-repset.yaml
+++ b/tyk-headless/templates/deployment-gw-repset.yaml
@@ -1,29 +1,16 @@
 apiVersion: apps/v1
-kind: {{ .Values.gateway.kind }}
+kind: StatefulSet
 metadata:
   name: gateway-{{ include "tyk-headless.fullname" . }}
   labels:
     app: gateway-{{ include "tyk-headless.fullname" . }}
     chart: {{ include "tyk-headless.chart" . }}
-    release: {{ .Release.Name }}
     heritage: {{ .Release.Service }}
 spec:
-{{- if eq .Values.gateway.kind "Deployment" }}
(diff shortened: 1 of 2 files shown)


abhishiekgowtham:
I'm not sure having STS by default is what everyone wants or deploying in their clusters. Having the ability to customise the resource kind is useful and leave the `type` to the administrators. Plus, not sure whether this PR has been tested and have tests to support it.

seanhoughton:
If you want to stick with a default behavior of ephemeral storage for persistent data that's up to your team. However, I would suggest making that extremely obvious in the documentation and the helm chart values.yaml comments.

The changes have been running fine on a K8S cluster. Do automated tests exist for these Helm charts?

PR #173, opened 2021-11-02
add-ability-to-modify-strategy

zalbiraw:
Add the ability to change the gateway strategy


letzya:
Should we document this option in the official docs and explain why

abhishiekgowtham (commented):
LGTM! 

Doesn't need to update `values.yaml`?

zalbiraw:
No this is optional. We can update it if we want.

PR #180, opened 2021-12-01
making user passed secrets mount in main container

cherrymu:
Making this minor change, so that the user passing mount values from values.yaml file will be correctly mounted under the main container rather the initContainer since it goes off once the main container is initialized and the mounted secret becomes null.


abhishiekgowtham (approved):
/lgtm

PR #184, opened 2022-02-01
[TD-597] Updated Ingress resource to support various Kubernetes versions

cherrymu:
Updated Ingress resource template to support latest Kubernetes API versions with backward compatibility also.


vesko-tyk (approved):
LGTM!

PR #197, opened 2022-04-14
TD-760 Add v4 support

zalbiraw:


tyk-headless/templates/deployment-pmp.yaml
```diff
@@ -88,37 +90,101 @@ spec:
             value: "4000"
           - name: TYK_PMP_PURGEDELAY
             value: "4"
-          {{- if .Values.mongo.enabled }}
-          - name: TYK_PMP_PUMPS_MONGO_TYPE
+{{ if .Values.backend }}
+{{ if eq .Values.backend "postgres" }}
```
zalbiraw, line 94: @vesko-tyk anyway I can simplify this? 

I tried `{{ if and .Values.backend (eq .Values.backend "postgres") }}` but it seems that the and does not short circuit and the second values causes a crash. I am trying to support older values.yamls

letzya: In go temple all parts are evaluated. You can create a function or leave it like that

tyk-pro/templates/deployment-dash.yaml
```diff
@@ -97,13 +97,30 @@ spec:
             value: "{{ default "false" .Values.redis.useSSL }}"
           - name: TYK_DB_REDISSSLINSECURESKIPVERIFY
             value: "true"
-          - name: TYK_DB_MONGOURL
+
+{{ if and .Values.backend (eq "postgres" .Values.backend) }}
```
letzya, line 101: Per comment above this needs to change to avoid crash

tyk-pro/values.yaml
```diff
@@ -76,6 +78,21 @@ mongo:
   # Default value: false
   useSSL: false
 
+postgres:
+  # Default value: tyk-postgres-postgresql.{{ .Release.Namespace }}.svc.cluster.local
```
letzya, line 82: It's a bit confusing that we write default values before the field name and also twice

letzya (dismissed):

tyk-pro/values.yaml
```diff
@@ -76,6 +78,21 @@ mongo:
...
+  # port: 5432
+  # Default value: postgres
+  # user: postgres
+  # password: 
+  # Default value: tyk_analytics
+  # database: tyk_analytics
+
+  # Enables SSL for MongoDB connection. MongoDB instance will have to support that.
+  # Default value: false
+  useSSL: false
```
drpebcak, line 94: Is this vestigial? It references mongo in the comment, and I don't see a reference to `postgres.useSSL` anywhere. Would be nice to be able to set the sslmode for the connection string.

tyk-headless/templates/deployment-pmp.yaml
zalbiraw: I see

tyk-pro/values.yaml
zalbiraw: Agreed

tyk-pro/values.yaml
zalbiraw: Yeah sorry copy paste error 😆

zalbiraw: @drpebcak if I am not mistaken you are referring to the `sslmode` option on the connection string right?

drpebcak: Yes, that's what I mean

zalbiraw: Makes perfect sense. I am thinking I should I just allow anyone to add whatever parameters to the postgres object in the values.yaml and then collapse that to create the connection string. That should give everyone the ability to do whatever they need to do.

tyk-headless/values.yaml
```diff
@@ -75,7 +71,16 @@ mongo:
...
+  useSSL: false
+
+# Postgres connection string parameters.
+postgres:
+  host: tyk-postgres-postgresql.tyk.svc.cluster.local
+  port: 5432
+  user: postgres
+  password:
+  database: tyk_analytics
+  sslmode:
```
zalbiraw, line 83: @drpebcak now you can pass any connection string parameters in there.

drpebcak:
After working through some other issues, this is working in my staging environment. What will it take to get this merged and released?

README.md
```diff
@@ -4,31 +4,42 @@ Tyk provides 3 different helm charts in this repo. Please visit the respective p
...
 
 *Important Note regarding MongoDB:* This helm chart enables the PodDisruptionBudget for MongoDB with an arbiter replica-count of 1.  If you intend to perform system maintenance on the node where the MongoDB pod is running and this maintenance requires for the node to be drained, this action will be prevented due the replica count being 1.  Increase the replica count in the helm chart deployment to a minimum of 2 to remedy this issue.
 
-Another option for Redis and MongoDB, if you want to get started quickly is to use our simple charts. **Please note that these provided charts must not ever be used in production and for anything but a quick start evaluation only, use external DBs or Official Helm charts for MongoDB and Redis in any other case.**
+Another option for Redis and MongoDB or PostgreSQL, if you want to get started quickly is to use our simple charts. **Please note that these provided charts must not ever be used in production and for anything but a quick start evaluation only, use external DBs or Official Helm charts for MongoDB or PostgreSQL and Redis in any other case.**
```
komalsukhani, line 31: I don't think we have simple charts for postgreSQL

zalbiraw: No we don't

komalsukhani: Should we remove PostgreSQL from that part then?

komalsukhani (approved):

buraksekili (approved):

Issue #178, opened 2021-11-30
Add runAsUser e fsGroup securityContext configuration


tauonico:
Fixed value in template for 'runAsUser' and 'fsGroup' cannot be validated against security context constraint on some environment.

Example:
```
securityContext:
        runAsUser: 1000
        fsGroup: 2000
```

To be:
```
securityContext:
        runAsUser: {{ .Values.securityContext.runAsUser }}
        fsGroup: {{ .Values.securityContext.fsGroup }}
```

buraksekili:
Hi @tauonico, we have an open ticket (#220) for `securityContext` configuration - hoping to release it soon.

buraksekili:
@tauonico we merged `securityContext` configuration #220 to the master. It will be part of the next release. Thanks!!

PR #200, opened 2022-06-03
Add openshift support

zalbiraw:
Also resolves https://github.com/TykTechnologies/tyk-helm-chart/issues/178

### DIFF
--- a/OpenShift.md
+++ b/OpenShift.md
@@ -1,0 +1,54 @@
+# Tyk Helm Chart - OpenShift
+To install Tyk using our helm chart on OpenShift please follow this guide:
+
+1. Create tyk namespace: `kubectl create ns tyk`
+
+2. Find the user and group UID range for the OpenShift project.
+```
+kubectl describe ns tyk | grep uid-range
+```
+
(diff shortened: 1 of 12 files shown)


tyk-pro/values.yaml
```diff
@@ -353,6 +362,9 @@ gateway:
   #  - secretName: chart-example-tls
   #    hosts:
   #      - chart-example.local
+  securityContext:
+    runAsUser: 1000
```
letzya, line 366: Sorry for my ignorance, what will the user's group be? (since we don't have `runAsGroup` set up here)

OpenShift.md
```diff
@@ -0,0 +1,55 @@
+# Tyk Helm Chart - OpenShift
```
letzya, line 1: @zalbiraw @craigtyk I am not sure this belongs in this repo. Shouldn't it be in tyk.io/docs? We can put it there under FAQ since we are not ready to officially support it?

zalbiraw: We can move this to docs and link it.

zalbiraw:
Will be add through a different PR.

PR #202, opened 2022-06-08
[TT-5524] Update bootstrap post-install hook and introduce new field in the values.yaml

buraksekili:
Current problem observed:
- `tyk-operator-conf` is not deleted if we do helm delete

- Failed Job doesn’t get deleted if we do helm delete. We end up with more jobs that fail

This PR adds 
- `before-hook-creation` to the delete policy of the post-install bootstrap. So that, jobs created by the bootstrap will be deleted either before a new hook is launched or successfully finished.
- new deletion hooks to clean up a secret of tyk-operator if it exists.
- new field called `operatorSecret` in the values.yaml. If it is set to true, `tyk-operator-conf` is created along with the dashboard bootstrap. Otherwise, it won't be created.

### DIFF
--- a/tyk-pro/templates/bootstrap-pre-delete.yaml
+++ b/tyk-pro/templates/bootstrap-pre-delete.yaml
@@ -1,0 +1,55 @@
+{{- if .Values.bootstrap -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: "bootstrap-pre-delete-{{ include "tyk-pro.fullname" . }}"
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+
+    # Removed the use of appVersion since it's not accurate. We'll put it back in v1.0
+    # when every app will have its own a chart
+    # app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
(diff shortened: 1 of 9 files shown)


komalsukhani (approved):

komalsukhani:
@buraksekili PR looks okay to me. I am not merging it yet since we plan to release it in next release.

andrei-tyk (approved):

komalsukhani (approved):

tyk-pro/templates/bootstrap-pre-delete.yaml
```diff
@@ -0,0 +1,55 @@
...
+    # Removed the use of appVersion since it's not accurate. We'll put it back in v1.0
+    # when every app will have its own a chart
+    # app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "-6"
```
komalsukhani, line 19: All looks good. Just curious to know why have to set weight to `-6`?

komalsukhani:
@buraksekili Please resolve the conflicts and then we can merge it.

tyk-pro/templates/bootstrap-pre-delete.yaml
buraksekili: My bad, I forgot to delete it. We can leave it with its default value since we have only one hook.

komalsukhani (approved):

PR #205, opened 2022-06-13
Bootstrap TLS fix

andrei-tyk:
This pull request enables the helm charts users to correctly deploy the tyk helm charts using TLS.

### DIFF
--- a/tyk-pro/templates/bootstrap-post-install.yaml
+++ b/tyk-pro/templates/bootstrap-post-install.yaml
@@ -76,10 +76,25 @@ spec:
           {{- end }}
           - name: TYK_DASHBOARD_DEPLOY
             value: dashboard-{{ include "tyk-pro.fullname" . }}
+          {{- if or .Values.dash.tls .Values.gateway.tls }}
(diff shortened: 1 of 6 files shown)


tyk-pro/templates/bootstrap-post-install.yaml
```diff
@@ -76,10 +76,25 @@ spec:
...
+            subPath: {{.Values.certificates.caBundleFileName}}
+            readOnly: false
+          {{- end }}
           - name: bootstrap-config-volume
             mountPath: /opt/scripts
       volumes:
+        {{- if or .Values.dash.tls .Values.gateway.tls }}
+        - name: ca-pemstore
+          configMap:
+            name: ca-pemstore
```
komalsukhani, line 96: Should we make this configurable?

andrei-tyk: I have thought about it too but my reasoning was that the client should have as fewer things  as possible to do to enable TLS and for it to work. If you consider we need this configurable, mark it as "needing change" and I will do it asap.

PR #206, opened 2022-06-16
[TT-5731] Added GithubAction to release helm chart

komalsukhani:
Added Github Action which release helm charts when a version tag is created.


buraksekili (approved):

vesko-tyk (approved):
Thanks @komalsukhani. LGTM!

PR #209, opened 2022-06-21
[TT-5529] Check for existing org

andrei-tyk:
Modification done to the bootstrapping script in order to check if an organisation with the same name/cname already exists in the database. Should an entry be detected, the client should be asked to remove it or disable bootstrapping.

### DIFF
--- a/tyk-pro/scripts/bootstrap.sh
+++ b/tyk-pro/scripts/bootstrap.sh
@@ -22,6 +22,31 @@ syntax_message(){
   exit 1
 }
 
+check_for_existing_organisations(){
+    EXISTING_ORGS=$(CURL_CA_BUNDLE=$CURL_CA_BUNDLE_TYK curl --silent --header "admin-auth: $TYK_ADMIN_SECRET" --header "Content-Type:application/json"  $DASHBOARD_HOSTNAME/admin/organisations 2>&1)
+    OWNER_NAMES=$(echo "$EXISTING_ORGS" | jq -r '.organisations | .[] | .owner_name | @base64')
+    if [[ -z "${OWNER_NAMES}" ]]
+    then
+      echo "No organisation present"
+    else
+      for row in $OWNER_NAMES; do
+        CURRENT_ORG_NAME=$(echo ${row} | base64 -d)
+        if [[ "$CURRENT_ORG_NAME" == "$TYK_ORG_NAME" ]]; then
+           echo "7"
+           return
+        fi
+      done
+      OWNERCNAMES=$(echo "$EXISTING_ORGS" | jq -r '.organisations | .[] | .cname | @base64')
+      for row in $OWNERCNAMES; do
+        CURRENT_ORG_CNAME=$(echo ${row} | base64 -d)
+        if [[ "$CURRENT_ORG_CNAME" == "$TYK_ORG_CNAME" ]]; then
+           echo "7"
+           return
+        fi
+      done
+    fi
+}
+
 generate_credentials(){
   error=0
   [[ -z "$1" ]] && return "6" || echo ""
@@ -56,6 +81,15 @@ exit_reason(){
       ;;
     "6")
       echo "ERROR: Hostname should not contain port number or protocol"
+      exit $1
+      ;;
+    "7")
+      echo "An organisation with this name already exists, please delete it or disable bootstrapping"
+      exit $1
+      ;;
+    "8")
+      echo "An organisation with this cname already exists, please delete it or disable bootstrapping"
+      exit $1
       ;;
     *)
       echo ""
@@ -157,6 +191,10 @@ log_ok(){
 
 if [ "$DASHBOARD_ENABLED" = "true" ] && [ "$BOOTSTRAP_DASHBOARD" = "true" ]
 then
+  echo "Checking if any organisations still exist"
+  ORG_EXISTS=$(check_for_existing_organisations)
+  exit_reason $ORG_EXISTS
+
   main $DASHBOARD_HOSTNAME
 fi
 


tyk-pro/scripts/bootstrap.sh
```diff
@@ -22,6 +22,31 @@ syntax_message(){
   exit 1
 }
 
+check_for_existing_organisations(){
+    EXISTING_ORGS=$(CURL_CA_BUNDLE=$CURL_CA_BUNDLE_TYK curl --silent --header "admin-auth: $TYK_ADMIN_SECRET" --header "Content-Type:application/json"  $DASHBOARD_HOSTNAME/admin/organisations 2>&1)
+    OWNER_NAMES=$(echo "$EXISTING_ORGS" | jq -r '.organisations | .[] | .owner_name | @base64')
```
buraksekili, line 27: Do we need to encode organisation name? Because at line 33, org_name is decoded again.

```suggestion
    OWNER_NAMES=$(echo "$EXISTING_ORGS" | jq -r '.organisations | .[] | .owner_name')
```

tyk-pro/scripts/bootstrap.sh
```diff
@@ -22,6 +22,31 @@ syntax_message(){
...
+        if [[ "$CURRENT_ORG_NAME" == "$TYK_ORG_NAME" ]]; then
+           echo "7"
+           return
+        fi
+      done
+      OWNERCNAMES=$(echo "$EXISTING_ORGS" | jq -r '.organisations | .[] | .cname | @base64')
+      for row in $OWNERCNAMES; do
+        CURRENT_ORG_CNAME=$(echo ${row} | base64 -d)
+        if [[ "$CURRENT_ORG_CNAME" == "$TYK_ORG_CNAME" ]]; then
+           echo "7"
```
buraksekili, line 43: Ignore me if i am wrong but I believe this should be `echo "8"` based on the `echo "8"`'s error message.
```suggestion
           echo "8"
```

tyk-pro/scripts/bootstrap.sh
```diff
@@ -22,6 +22,31 @@ syntax_message(){
...
+      echo "No organisation present"
+    else
+      for row in $OWNER_NAMES; do
+        CURRENT_ORG_NAME=$(echo ${row} | base64 -d)
+        if [[ "$CURRENT_ORG_NAME" == "$TYK_ORG_NAME" ]]; then
+           echo "7"
+           return
+        fi
+      done
+      OWNERCNAMES=$(echo "$EXISTING_ORGS" | jq -r '.organisations | .[] | .cname | @base64')
```
buraksekili, line 39: Unless I am missing some technical detail, we may skip encoding `OWNERCNAMES` because at line 41, we are decoding it again

tyk-pro/scripts/bootstrap.sh
```diff
@@ -56,6 +81,15 @@ exit_reason(){
...
     "6")
       echo "ERROR: Hostname should not contain port number or protocol"
+      exit $1
+      ;;
+    "7")
+      echo "An organisation with this name already exists, please delete it or disable bootstrapping"
+      exit $1
+      ;;
+    "8")
+      echo "An organisation with this cname already exists, please delete it or disable bootstrapping"
```
buraksekili, line 91: One minor thing. I think, it might be worth mentioning `deleting from Mongo/Postgre` instead of `please delete it`. But if you think the error message is clear enough, just keep it as it is. 😃

tyk-pro/scripts/bootstrap.sh
```diff
@@ -22,6 +22,31 @@ syntax_message(){
   exit 1
 }
 
+check_for_existing_organisations(){
+    EXISTING_ORGS=$(CURL_CA_BUNDLE=$CURL_CA_BUNDLE_TYK curl --silent --header "admin-auth: $TYK_ADMIN_SECRET" --header "Content-Type:application/json"  $DASHBOARD_HOSTNAME/admin/organisations 2>&1)
+    OWNER_NAMES=$(echo "$EXISTING_ORGS" | jq -r '.organisations | .[] | .owner_name | @base64')
+    if [[ -z "${OWNER_NAMES}" ]]
+    then
+      echo "No organisation present"
+    else
```
buraksekili, line 31: We may get rid of `else` statement if we return early after printing echo message.

tyk-pro/scripts/bootstrap.sh
andrei-tyk: Assume you have these two orgs: "Default org1" "Default org2"
If we wouldn't encode and decode it the space between the words "default" and "orgX" would cause our for loop to run 4 times instead of 2 times.

tyk-pro/scripts/bootstrap.sh
andrei-tyk: I believe having clarity in the code (especially bash code) is more important than getting rid of a single "else" statement. But I agree  with you that the statement can be deleted if we really wanted to.

tyk-pro/scripts/bootstrap.sh
andrei-tyk: Same as above.

tyk-pro/scripts/bootstrap.sh
andrei-tyk: I agree your version sounds more explicit.

andrei-tyk: Added in the next commit.

tyk-pro/scripts/bootstrap.sh
andrei-tyk: Fixed in the next commit.

andrei-tyk: Fixed.

tyk-pro/scripts/bootstrap.sh
andrei-tyk: Assume you have these two orgs: "Default org1" "Default org2"
If we wouldn't encode and decode it the space between the words "default" and "orgX" would cause our for loop to run 4 times instead of 2 times.

tyk-pro/scripts/bootstrap.sh
andrei-tyk: Fixed.

tyk-pro/scripts/bootstrap.sh
buraksekili: 🦸 makes sense. thanks!

buraksekili (approved):
🚀

tyk-pro/scripts/bootstrap.sh
komalsukhani: @andrei-tyk If you wrap whole expression inside [], it will return back an array.
So it will look like `jq -r ['.organisations | .[] | .owner_name']`

komalsukhani (approved):

PR #213, opened 2022-06-28
[TT-5609] Check Environment Configuration Before Bootstrapping

buraksekili:
## Description
This PR adds a validation on the environment configuration of the tyk-pro. 

- Checks if the DB License is empty, valid (of type `jwt`) or expired,
- Adds liveness check (using GW's `/hello` endpoint) for Gateway, Dashboard and Redis. If any of these are not ready, bootstrap won't start

Signed-off-by: Burak Sekili <buraksekili@gmail.com>

### DIFF
--- a/tyk-pro/scripts/init.sh
+++ b/tyk-pro/scripts/init.sh
@@ -36,28 +36,31 @@ wait_for_liveness () {
 
   while true
   do
-   attempt_count=$((attempt_count+1))
-  if [[ attempt_count -gt 100 ]]
-   then
-    echo "Failed to wait for liveness, exceeded attempt count"
-    exit 1
-  fi
+    attempt_count=$((attempt_count+1))
+    
+    # The timeout is set by default to 6 mins which is the default back-off limit for
+    # Kubernetes Jobs.
+    if [[ attempt_count -gt 180 ]]
+    then
+      echo "Failed to wait for liveness, exceeded attempt count"
+      exit 1
+    fi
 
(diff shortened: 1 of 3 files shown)


tyk-pro/scripts/init.sh
```diff
@@ -1,10 +1,75 @@
...
+
 set -e
+apk --no-cache add curl
+wait_for_liveness
 
-apk --no-cache add curl jq
 curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
 install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
 
 while [[ $(kubectl get pod -n $TYK_POD_NAMESPACE -l app=$TYK_DASHBOARD_DEPLOY -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}') != "True" ]]; do sleep 10; done && \
```
komalsukhani, line 74: @buraksekili Can we remove this check now?

buraksekili: yes @komalsukhani. since we are checking liveness with `wait_for_liveness` function, we do not need to check status of the dasboard pod again. I'll update accordingly 👍

buraksekili: updated @komalsukhani , thanks!!

komalsukhani (approved):

PR #243, opened 2022-11-06
[tyk-headless] Add persistent volume

cuttingedge1109:


cuttingedge1109:
@komalsukhani Did you have a chance to check this PR?

komalsukhani:
@cuttingedge1109 Can you fix following things
- Fix lint issues
- Add an entry in Changelog file
- Can you tell us how have you tested it?
- Can you add description above each new field you have added in values.yaml file?

buraksekili:
hi @cuttingedge1109, thanks for the contribution. Can you please update your PR according to @komalsukhani's feedback above? we are planning to move with this PR

tyk-headless/values.yaml
```diff
@@ -198,3 +198,10 @@ pump:
   mounts: []
 
 rbac: true
+
+persistence:
```
totogo, line 202: I would recommend putting this value under `gateway` since it is a pvc of `gateway`. Meanwhile renaming `templates/pvc.yaml` into `templates/pvc-gateway.yaml` to follow the previous convention of this repo and avoid confusion.

tyk-headless/values.yaml
```diff
@@ -139,6 +139,27 @@ gateway:
   #    hosts:
   #      - chart-example.local
 
+  ## Persistence parameters
+  ## ref: https://kubernetes.io/docs/user-guide/persistent-volumes/
+  ##
+  persistence:
+    ## Enable persistence on gateway using Persistent Volume Claims
+    ##
+    enabled: true
```
buraksekili, line 148: i guess it'd be better to disable persistence by default. if users need to enable it, they can set this field to `true`

caroltyk:
This issue should be resolved by https://github.com/TykTechnologies/tyk-helm-chart/pull/264

PR #250, opened 2022-11-30
[TT-7211] Fix headless templates

komalsukhani:
## Description

Validate `redis` and `mongo` fields are not nil before accessing it's child fields

## Related Issue

[TT-7211](https://tyktech.atlassian.net/browse/TT-7211)

## Test Coverage For This Change

Installation was successful with empty `mongo` and `redis` fields.

## Types of changesmpaty m

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). If PRing from your fork, don't come from your `master`!
- [x] Make sure you are making a pull request against our **`master` branch** (left side). Also, it would be best if you started *your change* off *our latest `master`*.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased](https://github.com/TykTechnologies/tyk-helm-chart/tree/HEAD)
 
+**Update:**
+- Check mongo and redis fields is not nil before using it's child fields
+
 ## [v0.13.0](https://github.com/TykTechnologies/tyk-helm-chart/tree/HEAD)
 
 [Full Changelog](https://github.com/TykTechnologies/tyk-helm-chart/compare/v0.12.0...v0.13.0)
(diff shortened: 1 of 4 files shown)


komalsukhani:
> It looks great considering the current problem that we faced. We also use `.Values.mongo` and `.Values.redis` in different places, for instance in [`tyk-headless/gw-deployment`](https://github.com/TykTechnologies/tyk-helm-chart/blob/388d56607702cd6ad994ed8c8e31266c140e4d46/tyk-headless/templates/deployment-gw-repset.yaml#L94). Should we add nil check to these as you did in the `helpers`? Or is it enough to add it to `helpers`?

Sorry I missed those cases. Will fix them

PR #258, opened 2023-01-23
[TT-6557] - Add Prometheus Pump configuration

buraksekili:
## Description
This PR updates Tyk Pump configuration in `tyk-pro` and `tyk-headless`. So that, from now on users can enable Prometheus pump from `values.yaml` file.

## Related Issue

https://tyktech.atlassian.net/browse/TT-6557
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). If PRing from your fork, don't come from your `master`!
- [ ] Make sure you are making a pull request against our **`master` branch** (left side). Also, it would be best if you started *your change* off *our latest `master`*.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

### DIFF
--- a/tyk-headless/values.yaml
+++ b/tyk-headless/values.yaml
@@ -74,8 +74,7 @@ redis:
   storage:
     database: 0
 
-# Choose the pumps backend for Tyk. [ "", "mongo", "postgres" ]
-backend: ""
+# mongo specifies MongoDB connection configuration for Tyk Pump if .Values.pump.backend is set to mongo.
 mongo:
   # The mongoURL value will allow you to set your MongoDB address.
   # Default value: mongodb://mongo.{{ .Release.Namespace }}.svc.cluster.local:27017/tyk_analytics
@@ -87,7 +86,7 @@ mongo:
   # Default value: false
   useSSL: false
 
-# Postgres connection string parameters.
+# postgres specifies PostgreSQL connection configuration for Tyk Pump if .Values.pump.backend is set to postgres.
 postgres:
   host: tyk-postgres-postgresql.tyk.svc.cluster.local
   port: 5432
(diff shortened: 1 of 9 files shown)


tyk-headless/values.yaml
```diff
@@ -194,6 +236,11 @@ pump:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+
+  # otherPumpBackend specifies if the backend apart from MongoDB, SQL or Prometheus is in use, e.g., Elasticsearch.
+  # Please set it to true while using other backends listed above. You can see all supported Pump backends here:
+  # https://tyk.io/docs/tyk-stack/tyk-pump/other-data-stores/
+  otherPumpBackend: false
```
komalsukhani, line 243: Should we remove the backend field from tyk-headless and instead use this field only to set pumpBackend?

tyk-headless/values.yaml
```diff
@@ -168,7 +173,44 @@ gateway:
...
+    customMetrics: ""
+    # By default, Pump creates a PodMonitor Custom Resource on your cluster if you use Prometheus Pump. Set .prometheusOperator.enabled to false
+    # for disabling installation of PodMonitor resource.
+    prometheusOperator:
+      # enabled determines whether the Prometheus Operator is in use or not. By default,
+      # it is enabled.
+      # Tyk Pump can be monitored with PodMonitor Custom Resource of Prometheus Operator.
+      # If enabled, PodMonitor resource is created based on .Values.pump.prometheusPump.prometheusOperator.podMonitorSelector
+      # for Tyk Pump.
+      enabled: true
```
komalsukhani, line 201: If we disable it, then will prometheus pump work?

buraksekili: ah, yes. it should be disabled. i forgot to switch off

tyk-headless/values.yaml
buraksekili: but then how can we understand the pump type? at the moment, we only set environment variables for SQL, MongoDB, and Prometheus pumps.

komalsukhani: Instead of backend we can use pumpBackend field.
In pro it makes sense to have backend so that based on the database dashboard is using we can create pump. But in headless we use it to create pump only.

